PR #30, opened 2014-03-12
Gitignore, and shinier "resolveURLTemplates"

shaqq:
Features:
- .gitignore file ignores node_modules and bower_components
- resolveURLTemplates is parameterized, so the macro values are not hard-coded like they are [here](https://github.com/dailymotion/vast-client-js/blob/master/src/util.coffee#L21)
- also adding another macro format, "%%MACRO_NAME%%", because urls may or may not come formatted
- added some test cases, all of which seem to pass

### DIFF
--- a/test/util.coffee
+++ b/test/util.coffee
@@ -17,3 +17,6 @@ describe 'VASTUtil', ->
 
         it 'should resolve playhead, with percents', ->
             resolve("http://test.com/%%CONTENTPLAYHEAD%%", CONTENTPLAYHEAD: 120).should.equal "http://test.com/120"
+
+        it 'should resolve weird cases', ->
+            resolve("http://test.com/%%CONTENTPLAYHEAD%%&[CONTENTPLAYHEAD]", CONTENTPLAYHEAD: 120).should.equal "http://test.com/120&120"


flasheye:
Hi!
> maybe adding a templating function 

That would be a great Feature! 
I have to add [RANDOM] (uppercase) and [TIMESTAMP].

Is this possible? Thanks!

michaelvial:
@flasheye I've done a PR https://github.com/dailymotion/vast-client-js/pull/140 for supporting theses missing macros. Can you tell me in the comments if this solution works for you?

flasheye:
@michaelvial Thanks for the very quick update! :) 
Unfortunately, I do not use coffee script, i just include the finish .js file.
When i have time i will try coffee ;)

michaelvial:
@flasheye Well, we've decided to merge it, theses macros are part of the spec (except for [RANDOM] but as we were handling [random] too, that wasn't a big change)
Anyway, we've released a new version (1.6.0) today that include the fix (in the bundle js file)

flasheye:
@michaelvial THANKS so much - it works 👍 
keep up the good work!

PR #64, opened 2015-01-06
Options parameter

mpotra:
+ Provides the optional `options` parameter to `.get()` methods.
Currently the only option available is `withCredentials` for XHR.

+ Provides support for `withCredentials` on XHR libraries, in order to allow browsers to pass cookies on.

Usage example: `vast.client.get(url, {'withCredentials': true}, cb)`

### DIFF
--- a/src/client.coffee
+++ b/src/client.coffee
@@ -6,8 +6,13 @@ class VASTClient
     @cappingMinimumTimeInterval: 0
     @timeout: 0
 
-    @get: (url, cb) ->
+    @get: (url, opts, cb) ->
         now = +new Date()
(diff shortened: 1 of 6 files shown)


hacfi:
:+1: Quite useful!

oleynikd:
Hi,

So if use `vast.client.get(url, {'withCredentials': true}, cb)` browser should pass previously received cookies?

mpotra:
@oleynikd Yes, it would include, in its request, any cookies set for a domain matching the `url` domain

oleynikd:
Great! Thanks

PR #65, opened 2015-01-12
fixed tracking event muted/unmuted

Lupennat:
changed to mute/unmute according to spec http://www.iab.net/media/file/VAST-2_0-FINAL.pdf


Lupennat:
I've no commit the updated vast-client.js and I've not update bower version.

I have to do it?

michaelvial:
Hi @Lupennat,
No you don't have to build `vast-client.js` nor update bower version. We handle this at each release.

PR #68, opened 2015-02-25
Make isURL more lenient to support IP addresses and localhost

adam-h:
Example URLS:
 - http://127.0.0.1:8080/second/wrapper_impression
 - http://localhost/test_impression


adam-h:
Also, it seems like only the Impression and Error URLs go through this. As all of the progress trackers were being fired correctly in my local environment.

PR #69, opened 2015-03-04
Http request timeout

Lupennat:
VASTClient passes timeout argument to VASTParser and URLHandler.

default timeout is 0

to set timeout:

DMVAST.client.timeout = 5000;

before calling

DMVAST.client.get( url, cb)

Added timeout also on test files


rs:
Please rebase your commit on master

PR #71, opened 2015-03-05
Http request timeout

Lupennat:
Added otions.timeout
DMVAST.client.get( url,options, cb)

options default is
{
    withCredentials : false,
    timeout : 0
}


Lupennat:
Can you update bower.json and vast-client.js?

dharFr:
I just tagged [version 1.3.3](https://github.com/dailymotion/vast-client-js/releases/tag/1.3.3). You should be able to use it with bower.

PR #72, opened 2015-03-10
+ Custom URLHandler and Extra CompanionAd Type Support

pcting:
- Add IFrameResource and HTMLResource companion support.
- Add ability to pass in a custom URLHandler as an options key so that
  you can do something like supply your own flash based XHR object.

### DIFF
--- a/src/parser.coffee
+++ b/src/parser.coffee
@@ -284,6 +284,12 @@ class VASTParser
             companionAd.id = companionResource.getAttribute("id") or null
             companionAd.width = companionResource.getAttribute("width")
             companionAd.height = companionResource.getAttribute("height")
+            for htmlElement in @childsByName(companionResource, "HTMLResource")
+                companionAd.type = htmlElement.getAttribute("creativeType") or 0
+                companionAd.htmlResource = @parseNodeText(htmlElement)
+            for iframeElement in @childsByName(companionResource, "IFrameResource")
+                companionAd.type = iframeElement.getAttribute("creativeType") or 0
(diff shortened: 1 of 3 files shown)


pcting:
If this is interesting for you guys to merge in, I can add in tests for it.

An example of the custom URLHandler using [CrossXHR](https://github.com/borisreitman/CrossXHR):

```javascript
var parseXml;

if (window.DOMParser) {
  parseXml = function(xmlStr) {
    return ( new window.DOMParser() ).parseFromString(xmlStr, "text/xml");
  };
} else if (typeof window.ActiveXObject != "undefined" && new window.ActiveXObject("Microsoft.XMLDOM")) {
  parseXml = function(xmlStr) {
    var xmlDoc = new window.ActiveXObject("Microsoft.XMLDOM");
    xmlDoc.async = "false";
    xmlDoc.loadXML(xmlStr);
    return xmlDoc;
  };
} else {
  parseXml = function() { return null; }
}

var swfUrlHandlerOptions = {
  urlhandler: {
    supported: function() {
      return !!CrossXHR;
    },

    get: function(url, options, cb) {
      try {
        var xhr = new CrossXHR();
        xhr.onreadystatechange = function() {
          if (xhr.readyState === 4) {
            cb(null, parseXml(xhr.responseText));
          }
        };
        console.debug('request to ', url);
        xhr.open('GET', url);
        xhr.send();
      } catch(e) {
        console.warn(e);
        cb();
      }
    }
  }
};

DMVAST.client.get(url, swfUrlHandlerOptions, function() { ... });
```

PR #75, opened 2015-05-27
Parse and store the AdParameters element for each linear creative

sclements:
Part of the VAST 2.0 spec and necessary for VPAID support.


rs:
Please edit the coffee script and not the compiled javascript.

sclements:
Updated with coffeescript

PR #77, opened 2015-06-10
override mime type

vforgione:
We were running into issues with an ad partner sending back valid VAST XML but with incorrect content types being set (they were `video/x-ms-asf`).

Adding `xhr.overrideMimeType('text/xml')` ensures that the returned data will be parsed by the following `xhr.responseXML`.


rs:
Please rebase so we don't get commits with the .idea stuff

PR #85, opened 2015-09-11
Add Companion Tests, Compiling vast-client.js, Updating Readme

spodlecki:
Few quick updates to the Companion parser. I noticed that master `.coffee` has what I needed (the IFrame and HTML Resources) -- so I just added quick tests to verify all works well. One note though, I do not like the type being passed as a `0`. Maybe we should be passing as text/iframe ?

* Added tests for IFrameResource
* Added tests for HTMLResource
* Updating README to include newly created types

Also, a note on compiling:
I was not able to run the bundle command, Browserify was not being nice. I ended up running:

```
npm run-script postinstall
browserify -s DMVAST dist/index.js -o vast-client.js
```

I'm going to be toying with the Non Linear scripts next as we'll probably be using this concept


test/parser.coffee
```diff
@@ -99,35 +99,72 @@ describe 'VASTParser', ->
 
             before (done) =>
                 companions = _response.ads[0].creatives[1]
+                console.log(companions)
```
rs: please remove this trace

src/parser.coffee
```diff
@@ -301,7 +301,7 @@ class VASTParser
             companionAd.width = companionResource.getAttribute("width")
             companionAd.height = companionResource.getAttribute("height")
             for htmlElement in @childsByName(companionResource, "HTMLResource")
-                companionAd.type = htmlElement.getAttribute("creativeType") or 0
+                companionAd.type = htmlElement.getAttribute("creativeType") or 'text/html'
```
rs: I'm not sure about this default and why it shouldn't be reflected for other type of resources (i.e: 3 lines below)?

spodlecki: From the looks of it, when declaring a `StaticResource` it is common place to apply a `creativeType` to the field... however, this does not seem to be the case with `HTMLResource` and `IFrameResource`

BrightCove: http://support.brightcove.com/en/video-cloud/docs/supported-vast-20-elements
```
[SUPPORTED]    <StaticResource creativeType="" />
[SUPPORTED]    <IFrameResource/>
[SUPPORTED]    <HTMLResource/>
```
VAST Spec: http://www.iab.net/media/file/VAST-2_0-FINAL.pdf
CompanionAds > HTMLResource and IFrameResource have no attributes

PR #91, opened 2015-09-30
Feature/inline extentions

### LINKED ISSUE #90: Support for parsing <Extensions> in <InLine>
We have a use case where we need to get at the <Extensions> values within <InLine>.  Seeing as by spec <Extensions> value can be any valid XML. I'm proposing a very generic parsed response. It'd be an array of extension elements each with name, value, attributes where they are string, string, object respectively. It'll be up to the end user to parse the value of each extension any farther as that would be custom code for the situation.

nerdynick:
Resolves #90

### DIFF
--- a/src/ad.coffee
+++ b/src/ad.coffee
@@ -4,5 +4,6 @@ class VASTAd
         @errorURLTemplates = []
         @impressionURLTemplates = []
         @creatives = []
+        @extensions = []
 
 module.exports = VASTAd
--- a/src/extension.coffee
+++ b/src/extension.coffee
@@ -1,7 +1,6 @@
 class VASTAdExtension
     constructor: ->
-        @name = null
         @attributes = {}
-        @value = null
+        @children = []
 
 module.exports = VASTAdExtension
--- a/src/extension.coffee
+++ b/src/extension.coffee
@@ -1,0 +1,7 @@
+class VASTAdExtension
+    constructor: ->
+        @name = null
+        @attributes = {}
+        @value = null
+
+module.exports = VASTAdExtension
--- a/src/extensionchild.coffee
+++ b/src/extensionchild.coffee
@@ -1,0 +1,7 @@
+class VASTAdExtensionChild
+    constructor: ->
+    	@name = null
+    	@value = null
+    	@attributes = {}
+
+module.exports = VASTAdExtensionChild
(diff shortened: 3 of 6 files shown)


rs:
Thanks for your contrib.

From the spec, the format should be:

``xml
<Extensions>
  <Extension type="some-type">
    <arbitrary>xml</arbitrary>
  </Extension>
</Extensions>
```

From what I read you missed the second level in your implementation.

**Comment by nerdynick** [comment]
Yea. I went just for the first children of Extensions because of inconstancy between vast 2 and 3. Vast 2.0 shows what you have there but Vast 3.0 just says it can be anything. But that was reading the PDF's I'll check the XSD to be sure.

**Comment by nerdynick** [comment]
XSD for 3.0 shows:
```
<xs:element name="Extensions" minOccurs="0" maxOccurs="1" type="Extensions_type"/>
```

Where 2.0 shows:
```
<xs:element name="Extensions" minOccurs="0" maxOccurs="1">
    <xs:complexType>
        <xs:sequence>
            <xs:element name="Extension" type="xs:anyType" minOccurs="0" maxOccurs="unbounded">
                <xs:annotation>
                    <xs:documentation>Any valid XML may be included in the Extensions node</xs:documentation>
                </xs:annotation>
            </xs:element>
        </xs:sequence>
    </xs:complexType>
</xs:element>
```

So what if I make the VastAd.extensions more of a collection supporting the 2.0 spec. Then add VastAd.extensionsRaw for the 3.0 spec. That way we have options.

**Comment by rs** [comment]
The "Human Readable VAST XML Schema" still show the sub element and the XSD tells the same:

```xml
<xs:complexType name="Extensions_type">
  <xs:sequence>
    <xs:element name="Extension" minOccurs="0" maxOccurs="unbounded">
      <xs:annotation>
        <xs:documentation>
          Any valid XML may be included in the Extensions node
        </xs:documentation>
      </xs:annotation>
      <xs:complexType>
        <xs:sequence>
          <xs:any minOccurs="0" maxOccurs="unbounded" processContents="lax" namespace="##any"/>
        </xs:sequence>
        <xs:anyAttribute namespace="##any" processContents="lax"/>
      </xs:complexType>
    </xs:element>
  </xs:sequence>
</xs:complexType>
```

nerdynick:
Ah, didn't notice that part. 

It also look like CreativeExtensions_type is of a similar structure. So what if I break the parse out into a function for just parsing all extension types.

Then modify the VastAdExtension to represent the <Extension> or <CreativeExtension> with an attribute collection like currently done and a child elements collection similar to what VastAdExtension currently looks like.

rs:
Sounds good

nerdynick:
This should be all updated now to match what we decided here.

PR #92, opened 2015-10-01
Add ability to pass the VAST response XML document in the options

rs:
In some cases, you may have the VAST response in hand, and you don't
want the VAST client to fetch it from an URL but still want to let it
resolve wrappers if any. This PR let you passe the XML VAST response
as the `response` field of the `options` parameter. In this case, the
`url` parameter is ignored and the provided response is returned to
the parser.

@dharFr: please review

### DIFF
--- a/src/urlhandler.coffee
+++ b/src/urlhandler.coffee
@@ -3,11 +3,15 @@ flash = require './urlhandlers/flash'
 
 class URLHandler
     @get: (url, options, cb) ->
+        # Allow skip of the options param
         if not cb
             cb = options if typeof options is 'function'
             options = {}
 
-        if options.urlhandler && options.urlhandler.supported()
+        if options.response?
+            # Trick: the VAST response XML document is passed as an option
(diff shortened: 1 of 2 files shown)


src/urlhandler.coffee
```diff
@@ -3,11 +3,15 @@ flash = require './urlhandlers/flash'
...
+        # Allow skip of the options param
         if not cb
             cb = options if typeof options is 'function'
             options = {}
 
-        if options.urlhandler && options.urlhandler.supported()
+        if options.response?
+            # Trick: the VAST response XML document is passed as an option
+            cb(null, options.response)
+        else if options.urlhandler && options.urlhandler.supported()
```
dharFr: Nothing new, but this line could be simplified to the following: 
``` coffeescript
else if options.urlhandler?.supported()
```

dharFr:
The code looks good to me. 
But I think you should add a unit-test to cover that special case before merging it.

PR #93, opened 2015-10-03
Add support for protocol agnostic urls to parser

hydrozen:
ex: //www.example.com


hydrozen:
I’m not exactly sure how to add a test for this.

PR #102, opened 2016-03-02
Implementation for nonlinears. Working for Wrapper and Inline.

ufucuk:
I was not able squash the commits that is why I created a new pull request.


ufucuk:
Looks good now? :)

PR #106, opened 2016-07-21
fix: coffescript and compiled files with ie9 setAttribute -> id

rubeniskov:


rubeniskov:
http://www.w3schools.com/xml/met_element_setattribute.asp

michaelvial:
Hi @rubeniskov

You're right, better use `setAttribute()` here. Can you :
- Add a condition to handle only `Wrapper` and `Inline` elements so we can get rid off `adTypeElement.id = adElement.getAttribute("id");`
- Squash commits into a single one
- Do not change the `vast-client.js` file. I'll rebuild it once your PR is merged

Thx

rubeniskov:
Ouch, sorry about that! for the make changes in vast-client.js!

Thanks for replay!!

michaelvial:
Will be fixed in https://github.com/dailymotion/vast-client-js/pull/111

PR #108, opened 2016-08-03
Added support for CompanionClickTracking property to VAST parser

TLFyber:
When parsing companion ads, the VAST parser currently does not
load the optional CompanionClickTracking parameter. The value of
this parameter is required when implementing click tracking, so it
is worth including it in the parser's output. I added a line to read
the value of this parameter and return it in the companion ad object


michaelvial:
Thanks @TLFyber !

PR #109, opened 2016-08-03
Added support for multiple companion click tracking properties

TLFyber:
I have to apologise for an oversight on my part. Earlier I [submitted a PR](https://github.com/dailymotion/vast-client-js/pull/108) which added support for the CompanionClickTracking property which appears in VAST companion ads. The VAST spec doesn't make it entirely clear, but companion ads can have multiple CompanionClickTracking elements, indicating that all of those URLs should be pinged if the companion ad is clicked.

I tried to find a better source of information to confirm the expected behaviour but the most I could discover is that several VAST tags in the wild appear to be using multiple CompanionClickTracking properties. I therefore modified my original change to the parser so that it returns an array called companionClickTrackingURLTemplates instead of a string called companionClickTrackingURLTemplate.

Sorry for any inconvenience this caused :( I'm finding it quite difficult to get reliable information on the expected format of VAST tags.

### DIFF
--- a/src/parser.coffee
+++ b/src/parser.coffee
@@ -369,6 +369,7 @@ class VASTParser
             companionAd.id = companionResource.getAttribute("id") or null
             companionAd.width = companionResource.getAttribute("width")
             companionAd.height = companionResource.getAttribute("height")
+            companionAd.companionClickTrackingURLTemplates = []
             for htmlElement in @childsByName(companionResource, "HTMLResource")
                 companionAd.type = htmlElement.getAttribute("creativeType") or 'text/html'
                 companionAd.htmlResource = @parseNodeText(htmlElement)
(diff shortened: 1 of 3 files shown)


michaelvial:
Sorry for the late reply @TLFyber but after thinking about this, I can't merge this PR as it is. This causes a breaking change as `companion.companionClickTrackingURLTemplate` property doesn't exist anymore.

We've got two options:
- We still set `companion.companionClickTrackingURLTemplate` property so there will be no breaking changes and I can merge this PR.  And we'll delete this property in the next major version.
- We keep this PR as it is and merge it when we'll release the next major version.

TLFyber:
Sorry it also took me a long time to get back to you @michaelvial, fair point about the breaking change, I took the first option you suggested and reinstated the companionClickTrackingURLTemplate property so it sets that property to the first URL it finds and then stores all the urls it finds in the companionClickTrackingURLTemplates array. Let me know if it looks ok or if there's further changes required, cheers

PR #112, opened 2016-09-14
First onreadystatechange then send

ksyrytczyk:
If nested URL inside `VASTAdTagURI` tag is `http://`, `xhr.send()` will throw an error and a `try` block will end without calling callback.


michaelvial:
Thx for fixing this

piecioshka:
@ksyrytczyk thanks a lot!

PR #119, opened 2016-10-12
Ping <Error> when no ads found

johansatge:
Hi,

This PR is about pinging `<Error>` URLs when no ads are found in the given VAST.

Currently, we do that by looking for `<Ad>` nodes in the response ([here](https://github.com/dailymotion/vast-client-js/blob/2a64c454b618a2f9eb4e5c79b9e58ba62a9ba682/src/parser.coffee#L78-L89)).
Instead, we could look for `<Ad>` nodes with a `<Creative>`.

With this method, the `<Error>` URLs would still be pinged in case of a VAST with an empty `<Ad>`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<VAST version="3.0">
    <Ad>
        <InLine>
        </InLine>
    </Ad>
    <Error>
        [...]
    </Error>
</VAST>
```

### DIFF
--- a/src/parser.coffee
+++ b/src/parser.coffee
@@ -77,12 +77,12 @@ class VASTParser
 
             complete = (errorAlreadyRaised = false) =>
                 return unless response
-                hasAdWithCreative = false
+                noCreatives = true
                 for ad in response.ads
                     return if ad.nextWrapperURL?
                     if ad.creatives.length > 0
-                        hasAdWithCreative = true
-                if not hasAdWithCreative
+                        noCreatives = false
+                if noCreatives
                     # No Ad Response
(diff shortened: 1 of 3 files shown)


src/parser.coffee
```diff
@@ -77,14 +77,18 @@ class VASTParser
 
             complete = (errorAlreadyRaised = false) =>
                 return unless response
+                hasAdWithCreative = false
```
michaelvial: I would prefer to name it `noCreatives`

johansatge: I agree, it makes the `if noCreatives` condition more readable 👍

michaelvial: thx

PR #120, opened 2016-10-13
Improve error handling system

michaelvial:
I've added new errors through the code so we should better know what happened when calling `DMVAST.client.get()`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,19 +1,62 @@
 # VAST Parser
 The VAST Parser is accessed through *DMVAST.client* object.
 
-### get( url [, options ], success )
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -1,19 +1,62 @@
...
+// ...
+// 4 minutes later
+
+// Work
+DMVAST.client.get(VASTUrl, cb);
+```
+
+## Method
+
 ### get( url [, options ], success )
```
dharFr: This line should be `### get( url [, options ], done )`, right?

michaelvial: oops

Issue #123, opened 2016-10-24
Wrapper URLs not called when passing in Response Document


rizhang:
I have a situation where I am passing in the response document and not passing a URL. When the response document has a `<Wrapper>` object, the `<Wrapper>`'s URL is not called, and instead the original document I passed in is re-used.

Suggested Fix is to not re-use the `response` object on subsequent calls for `<Wrapper>` objects.

Steps to Reproduce:

* XML Document

```
<?xml version="1.0" encoding="UTF-8"?>
<VAST xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="vast.xsd" version="2.0">
   <Ad id="27209779">
      <Wrapper>
         <AdSystem version="1.0">adnxs</AdSystem>
         <VASTAdTagURI>http://bs.serving-sys.com/BurstingPipe/adServer.bs?cn=is&amp;c=23&amp;pl=VAST&amp;pli=12627946&amp;PluID=0&amp;pos=6644&amp;ord=[timestamp]&amp;cim=1</VASTAdTagURI>
         <Impression>http://ams1-ib.adnxs.com/it?e=wqT_3QLdAXTdAAAAAwDWAAUBCAAQABi7xP7klOiglHggACotCQANAQARDQgEABkRCQAhEQkAKREJ8JMwATgAQABIAFCz4PwMWABgAGgAeACAAQCYAQCgAQCoAQCwAQC4AQPAAQXIAQDQAQDYAQDgAQHwAQDYAgDgAgCAAwCIAwGQAwCYAwCgAwCqAwDAA6wCyAMA2AMA4AMA6AMA-AMAgAQAkgQDL2NymAQAqAQAsgQKCAAQABgAIAAwALgEAMAEAMgEANoEAggA4AQA8AQA&amp;s=4adc7b3cf7f16d3046ebc86d8ff055112a36cc7a</Impression>
         <Creatives>
            <Creative AdID="27209779">
               <Linear>
                  <VideoClicks>
[...]

PR #125, opened 2016-10-25
only call error callback after request complete and status is not 200

strworkstation:
#123

Error callback was called when readyStatus was 1 (OPENED), 2 (HEADERS_RECEIVED), or 3(LOADING), which caused the response to become null when it shouldn't.


michaelvial:
@strworkstation Thx for fixing this!

PR #126, opened 2016-10-27
Retrieve all Ad/Inline elements values and attributes

michaelvial:
Update the VAST Parser so we can retrieve:
- [x] All `<Inline>` element attributes/nodes
- [x] `<Icons>` element


dharFr:
Good to be merged 👍

PR #129, opened 2016-11-07
Use wrapper videoClickThru when Inline videoClickThru is null

michaelvial:
According to the [VAST 2.0 spec](http://www.iab.com/wp-content/uploads/2015/11/VAST-2_0-FINAL.pdf#page=14), the `Wrapper` element can contains an `videoClicks/clickThrough` URL. It should be use to replace any `Inline` ad that doesn't have a `videoClicks/clickThrough` URL.

Note: In VAST 3.0 and 4.0 versions, the wrapper element can't provide a `videoClicks/clickThrough` URL. However, this "feature" will still work.


test/parser.coffee
```diff
@@ -37,228 +37,253 @@ describe 'VASTParser', ->
...
 
         it 'should have merged top level error URLs', =>
-            @response.errorURLTemplates.should.eql ["http://example.com/wrapper-error", "http://example.com/error"]
+            @response.errorURLTemplates.should.eql ["http://example.com/wrapperA-error", "http://example.com/wrapperB-error", "http://example.com/error"]
 
         describe '#For the 1st ad', ->
+            ad1 = null
+
+            before (done) =>
+                ad1 = _response.ads[0]
```
dharFr: You should reset `ad1` to `null` in an `after()` hook to avoid potential side-effect on upcoming test suites. 
(+ I think the `done()` function should only be used when the hook includes asynchronous calls)

test/parser.coffee
```diff
@@ -37,228 +37,253 @@ describe 'VASTParser', ->
...
-            it 'should have 2 impression URLs', =>
-                _response.ads[1].impressionURLTemplates.should.eql ["http://example.com/wrapper-impression", "http://127.0.0.1:8080/second/wrapper_impression", "http://example.com/impression1"]
+            #Linear
+            describe '1st creative (Linear)', ->
+                linear = null
 
-            it 'should have 1 creative', =>
-                _response.ads[1].creatives.should.have.length 1
+                before (done) =>
+                    linear = _response.ads[0].creatives[0]
```
dharFr: Same here. I'd recommend to reset `linear` to `null` in an `after()` hook.

test/parser.coffee
```diff
@@ -37,228 +37,253 @@ describe 'VASTParser', ->
...
+                    icon.iconClickThroughURLTemplate.should.equal "http://example.com/linear-clickthrough"
+                    icon.iconClickTrackingURLTemplates.should.eql ["http://example.com/linear-clicktracking1", "http://example.com/linear-clicktracking2"]
+                    icon.iconViewTrackingURLTemplate.should.equal "http://example.com/linear-viewtracking"
+
+            #Companions
+            describe '2nd creative (Companions)', ->
+                companions = null
+
+                before (done) =>
+                    companions = _response.ads[0].creatives[1]
```
dharFr: Again ^^. `after()` hook. 😉 
The same pattern is used several times bellow. Please make sure to fix it everywhere. 
Thx.

PR #134, opened 2016-11-18
Update xmlhttprequest.coffee

injonsed:
Added async option for XMLHttpRequest.open

### DIFF
--- a/src/urlhandlers/xmlhttprequest.coffee
+++ b/src/urlhandlers/xmlhttprequest.coffee
@@ -12,8 +12,10 @@ class XHRURLHandler
             return cb(new Error('XHRURLHandler: Cannot go from HTTPS to HTTP.'))
 
         try
+            async = options.async or true
+
             xhr = @xhr()
-            xhr.open('GET', url)
+            xhr.open('GET', url, async)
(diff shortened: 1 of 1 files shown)


michaelvial:
Hi @injonsed and sorry for the late reply

I'm reluctant for adding this option, having the need for synchronous Xhr requests is really, really uncommon. But if you really need to, you can use a custom `URLHandler` to override the one provided by the VAST Client. We haven't done a proper documentation about how to create a `URLHandler`, but here is a brief example:

```javascript
// Sync Ajax request URLHandler
const customUrlHandler = function() {
  return {
    supported() { return true },
    get(url, options, cb) {
      $.ajax({
        url: url,
        async: false,
        dataType : 'xml',
      })
      .done((xml) => cb(null, xml))
      .fail((err) => cb(new Error(`Could not resolve VAST Wrapper [${url}][status:${err.status}]`)))
    }
  }
}

// Fetching vast.xml with customURLHandler
DMVAST.client.get('http://example.dailymotion.com/vast.xml', { urlhandler: customUrlHandler }, function(response, error)
{
  // process the VAST response
});
```

injonsed:
@michaelvial thanks you for answer.

PR #138, opened 2016-12-07
Add Support For Expand & Collapse events to Tracker

nuclearghost:
These events are part of the VAST 2.0 spec, but were not currently
supported by the tracker. This adds the ability to fire these events by
calling the `setExpand` method with a boolean value for expanded or
collpased


michaelvial:
Thx @nuclearghost for the PR

PR #140, opened 2016-12-12
Macros `RANDOM`, `TIMESTAMP` and `ASSETURI`

michaelvial:
Add support for macros `RANDOM`, `TIMESTAMP` and `ASSETURI`  in the VAST Client.

### DIFF
--- a/src/util.coffee
+++ b/src/util.coffee
@@ -19,7 +19,7 @@ class VASTUtil
         variables["ERRORCODE"] = 900 if variables["ERRORCODE"]? and not /^[0-9]{3}$/.test(variables["ERRORCODE"])
 
         # Calc random/time based macros
-        variables["CACHEBUSTING"] = Math.round(Math.random() * 1.0e+8)
+        variables["CACHEBUSTING"] = @leftpad(Math.round(Math.random() * 1.0e+8).toString())
         variables["TIMESTAMP"]    = @encodeURIComponentRFC3986((new Date).toISOString())
 
         # RANDOM/random is not defined in VAST 3/4 as a valid macro tho it's used by some adServer (Auditude)
@@ -42,6 +42,9 @@ class VASTUtil
         return encodeURIComponent(str).replace(/[!'()*]/g, (c) ->
             return '%' + c.charCodeAt(0).toString(16)
         )
+
+    @leftpad: (str) ->
+        return if str.length < 8 then ('0' for [0...8-str.length]).join('') + str else str
 
     @storage: do () ->
         try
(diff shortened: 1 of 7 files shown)


src/util.coffee
```diff
@@ -8,13 +8,15 @@ class VASTUtil
...
+    @resolveURLTemplates: (URLTemplates, variables = {}) ->
         URLs = []
 
-        variables ?= {} # ["CACHEBUSTING", "random", "CONTENTPLAYHEAD", "ASSETURI", "ERRORCODE"]
-        unless "CACHEBUSTING" of variables
-            variables["CACHEBUSTING"] = Math.round(Math.random() * 1.0e+10)
-        variables["random"] = variables["CACHEBUSTING"] # synonym for Auditude macro
+        variables["ASSETURI"]     = @encodeURIComponentRFC3986(variables["ASSETURI"]) if variables["ASSETURI"]?
+        variables["CACHEBUSTING"] = Math.round(Math.random() * 1.0e+10) unless variables["CACHEBUSTING"]?
+        variables["TIMESTAMP"]    = (new Date).toISOString() unless variables["TIMESTAMP"]?
```
dharFr: As the `CACHEBUSTING`, `TIMESTAMP`, `random`, `RANDOM` and `ERRORCODE` variable might be defined outside of this function, and then might have an arbitrary value, it might be a good idea to `@encodeURIComponentRFC3986()` all of them. What do you think?

dharFr: Same for `CONTENTPLAYHEAD` btw.

src/util.coffee
```diff
@@ -11,9 +11,16 @@ class VASTUtil
...
-        variables["TIMESTAMP"]    = (new Date).toISOString() unless variables["TIMESTAMP"]?
+        # Encode String variables, when given
+        variables["ASSETURI"]        = @encodeURIComponentRFC3986(variables["ASSETURI"]) if variables["ASSETURI"]?
+        variables["CONTENTPLAYHEAD"] = @encodeURIComponentRFC3986(variables["CONTENTPLAYHEAD"]) if variables["CONTENTPLAYHEAD"]?
+
+        # Set default value for invalid ERRORCODE
+        variables["ERRORCODE"] = 900 if variables["ERRORCODE"]? and not /^[0-9]{3}$/.test(variables["ERRORCODE"])
+
+        # Calc random/time based macros
+        variables["CACHEBUSTING"] = Math.round(Math.random() * 1.0e+8)
```
dharFr: If we really want to comply with the spec, we probably need to left-pad this value with `0`s...  :confounded:

PR #142, opened 2016-12-14
Remove empty #text/ #comments node from parsed extensions

michaelvial:
While parsing `<Extensions>` nodes,  the parser was creating an `extension` object for every `#comment` or empty `#text` nodes found.

They are now ignored an only non-empty `#text` node or `xml` node are saved.

### DIFF
--- a/test/parser.coffee
+++ b/test/parser.coffee
@@ -93,7 +93,7 @@ describe 'VASTParser', ->
             it 'validate third extension', =>
                 ad1.extensions[2].attributes['type'].should.eql "Count"
                 ad1.extensions[2].children.should.have.length 1
-                ad1.extensions[2].children[0].name.should.eql "total_available"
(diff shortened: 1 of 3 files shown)


test/parser.coffee
```diff
@@ -93,7 +93,7 @@ describe 'VASTParser', ->
             it 'validate third extension', =>
                 ad1.extensions[2].attributes['type'].should.eql "Count"
                 ad1.extensions[2].children.should.have.length 1
-                ad1.extensions[2].children[0].name.should.eql "total_available"
+                ad1.extensions[2].children[0].name.should.eql "#cdata-section"
```
dharFr: Could you add an additional unit-test for the following test-case? 
```
<Extension type="nocddata">some text without CDDATA</Extension>
```

PR #147, opened 2016-12-15
Parse all nonlinear attributes

michaelvial:
Parse and return `<nonlinear>` attributes:
- expandedWidth
- expandedHeight
- scalable
- maintainAspectRatio

Also convert `<nonlinear>` `minSuggestedDuration` value to second (as we do with `<linear><duration>` value).


johansatge:
Works like a charm, let's merge.

image

PR #159, opened 2017-04-13
Return creative id, adId, sequence and apiFramework attributes

michaelvial:
The VAST parser now returns Creative `id`, `adId`, `sequence` and `apiFramework` attributes.


johansatge:
Looks good 👌  let's merge.

image

Issue #79, opened 2015-07-01
Architecture issue


StevenLangbroek:
Hey guys,

I'm trying to use this in a componentised way (we implement this in our platform, but I'm building a standalone debugger). What I want to do should be relatively straight forward if it wasn't so tightly coupled:
- Show choice: `By URL` or `By XML`
- If `By XML`, show CodeMirror editor and have our employees input XML
- If `By URL`, run through `VASTClient.client`.

The problem here is when users choose `By XML`. Because of the way the client is structured, there's no way to just parse XML to the Object literal format you use, without invoking `URLHandler.get`. This seems like "bad architecture". The XML parser should be a static method on `Parser`, and be invokable without needing to call `URLHandler.get`. Would you be interested in a PR for this?

Thanks!

rs:
Yes it would be interested to decouple the URL handler. Feel free to PR.

PR #175, opened 2017-07-03
Add VASTParser.load() to allow initial VAST to be loaded as XML document

yowcow:
Related issue: #79

Changes
-------

+ Add a new method `load()` to load initial VAST as XML document
+ Add a new method `parseXmlDocument()` that parses XML document
+ Modify existing method `_parse()` to use `parseXmlDocument` on parsing XML document fetched through XHR

Motivation
----------

An application may already have a VAST XML document available within its code, not just having a URL to fetch VAST.

It's handy to have a way to inject VAST wrapper document into a parser, letting existing implementations complete gathering VASTs.

This feature will work perfectly in the project I'm currently working on.

### DIFF
--- a/src/parser.coffee
+++ b/src/parser.coffee
@@ -32,6 +32,13 @@ class VASTParser
         @_parse url, null, options, (err, response) ->
             cb(response, err)
 
+    @load: (xml, options, cb) ->
+        if not cb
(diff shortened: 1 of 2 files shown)


michaelvial:
Thanks a lot @yowcow!

PR #179, opened 2017-07-13
Fix duration parser

VanCyric:


michaelvial:
Well spotted @VanCyric. Thx.

PR #188, opened 2017-09-02
Return request object from urlhandlers

chrishoage:
When unmounting a component that makes a request we need to be able to abort the request. Coffeescript has an implicit return on the last line, and was returning the result from the send method for the xhr request. 

This instead returns the xhr request so that anything calling `client.get` is free to `xhr.abort()` the request preventing the callback from firing. 

I took a stab at the flash and node handlers, however I wasn't sure if the same approach was appropriate for those. Let me know if its not and I'll back those changes out. 

There should be an ability to abort xhr requests if necessary, currently that ability does not exist.

### DIFF
--- a/src/urlhandlers/flash.coffee
+++ b/src/urlhandlers/flash.coffee
@@ -24,4 +24,6 @@ class FlashURLHandler
              xmlDocument.loadXML(xdr.responseText)
              cb(null, xmlDocument)
 
+        return xdr
+
 module.exports = FlashURLHandler
--- a/src/urlhandlers/node.coffee
+++ b/src/urlhandlers/node.coffee
@@ -37,4 +37,6 @@ class NodeURLHandler
             fn = timeout_wrapper req
             timing = setTimeout( fn, options.timeout or 120000 );
 
+            return req;
+
 module.exports = NodeURLHandler
--- a/src/urlhandlers/xmlhttprequest.coffee
+++ b/src/urlhandlers/xmlhttprequest.coffee
@@ -24,6 +24,7 @@ class XHRURLHandler
                     else
                         cb(new Error("XHRURLHandler: #{xhr.statusText}"))
             xhr.send()
+            return xhr;
         catch
             cb(new Error('XHRURLHandler: Unexpected error'))
 


michaelvial:
Hi @chrishoage 

A new `client.cancel` method would make sense. This should abort any pending request and execute the `client.get `callback with a new error (something like `canceled by user`).

Or were you thinking of another use case?

chrishoage:
Why would the get callback need to run at all? 

The case I'm describing is a component that unmounts before the callback is fired. We do not want to fire the callback at all in that case. 

calling `xhr.abort()` prevents the callback from firing.

chrishoage:
Maybe I can describe the case better. 

We have a React component `VastPlayer` which uses `vast-client`. 

Lets say `VastPlayer` is unmounted while the request is in flight. What we want is to be able to directly call `xhr.abort()` on the request so that the callback isn't fired inside an unmounted component.

michaelvial:
This surely deserve a method in the `client` class to abort the VAST fetching/parsing. This method could take an argument to do track or not the interruption (some ad servers really want to know what happen, let's say they deliver 100 inventory (VAST response), they want to get the same number in impressions/errors).

In the mean time, you might achieve what you want using a `customUrlResolver`. This way you'll be able to abort the request on your side.

Something like:
```js
var customUrlHandler = {
  abort: function() {
    this.xhr.abort()
  },
  xhr: new window.XMLHttpRequest(),
  supported: function() { return true },
  get: function(url, options, cb) {
    // making the VAST request
    this.xhr.open('GET', url)
   // see https://github.com/dailymotion/vast-client-js/blob/master/src/urlhandlers/xmlhttprequest.coffee for full code
  }
}

var options = { urlhandler: customUrlHandler }

// Fetching/Parsing the VAST
DMVAST.client.get('http://example.dailymotion.com/vast.xml', options, function(response, error)
{
  // process the VAST response
});

// Abort xhr request when unmounting the component
onUnmount = function() {
  customUrlHandler.abort()
}
```

chrishoage:
Sure, I can expose this as an method on the `client` class. 

> This method could take an argument to do track or not the interruption (some ad servers really want to know what happen, let's say they deliver 100 inventory (VAST response), they want to get the same number in impressions/errors).

What might this look like?

eliamaino-fp:
Hey @chrishoage I will close this PR since it became quite outdated and after the `2.0` release we moved completely to es6. If you still want to implement a way to cancel fetching or parsing requests, take a look at the `2.0` and feel free to open a new PR in case :)

Issue #165, opened 2017-04-21
Try to load VAST URL over HTTPS if location is HTTPS


apptv:
Thanks for this great library.

I would like to propose an improvement for the Ad Tag Waterfalling.

In case the Ad Player is integrated on a secure site and one of the Ad Tags in the waterfall is pointing to a non secure site, the complete process is stopped.

```
XHRURLHandler.get = function(url, options, cb) {
    var xhr;
    if (window.location.protocol === 'https:' && url.indexOf('http://') === 0) {
      return cb(new Error('XHRURLHandler: Cannot go from HTTPS to HTTP.'));
    }

```

Because most AdServes support http and https, the Handler should try to fix the configuration error and load the URL via https instead of stopping.

michaelvial:
@apptv 

You can use a custom urlHandler as describe [in the doc](https://github.com/dailymotion/vast-client-js/blob/master/docs/api.md#get-url--options--done-)

Here is a quick example:
``` javascript
var urlhandler = {
  supported: function() { return true },
  get: function(url, options, cb) {
    if (window.location.protocol === 'https:' && url.indexOf('http://') === 0) {
      // Unsecure request - let's try https
      url = url.replace(/http/, "https")
    }
    //...
  }
}

var options = { urlhandler: urlHandler }

DMVAST.client.get('http://example.dailymotion.com/vast.xml', options, function(response, error)
{
  // process the VAST response
});
```

Hope it helps

apptv:
@michaelvial 
Thanks a lot for the quick reply and solution. Obviously I missed this part in the docs.

Just to mention it. If the Ad Player lives in a friendly iFrame.
`<iframe src='about:blank'></iframe>`
window.location.protocol returns 'about', so the check for https will always fail.

michaelvial:
@apptv If you find a good solution for this `https` try, don't hesitate to do a PR :)

apptv:
@michaelvial 
Sorry, but I am not used to coffee script.

This is my solution in javascript.

```
  var myXHRURLHandler = {
    'tryHttps' : function(url) {
      var protocol = window.location.protocol;
      if (protocol === 'about:') {
        if(document.referrer) {
          protocol = document.referrer.split('://')[0] + ':';
        } else {
          protocol = 'https:';
        }
      }
      if (protocol === 'https:' && url.search(/^http:\/\//i) === 0) {
        //Try to fetch via https
        url = url.replace(/^http:\/\//i, 'https://');
      }
[...]

PR #192, opened 2018-01-03
added error message for empty or malformed xml response

jgainfort:
For QOS Purposes I needed a way to distinguish between Invalid Vast Responses and Empty or Malformed XML Responses. This PR simply changes the error message on such occasions. Not sure if handy to anyone but myself, but figured its worth a PR just in case.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1433 @@
+{
+  "name": "vast-client",
+  "version": "1.8.0",
+  "lockfileVersion": 1,
+  "requires": true,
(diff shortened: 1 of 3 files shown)


michaelvial:
@jgainfort you pass a custom `urlHandler` to the vastClient `get()` method. This allows you to redefine the error messages.

See issue : https://github.com/dailymotion/vast-client-js/issues/165

eliamaino-fp:
See issue #165

Issue #193, opened 2018-01-24
callback of DMVAST.parser.load is called with params in wrong order


Knotschi:
The documentation says:
```
DMVAST.parser.load(xml, options, function(response, error) {
  // process the VAST response
});
```

But in the code the callback is called in opposite order: [`cb(error, response)`](https://github.com/dailymotion/vast-client-js/blob/master/src/parser.coffee#L106)

[even the test is using the order of first error than response:](https://github.com/dailymotion/vast-client-js/blob/master/test/parser.coffee#L547-L549)
```
VASTParser.load xml, (err, response) =>
    @response = response
    done()
```

[For DMVAST.parser.parse on the other side the order seems to be explicitly twisted:](https://github.com/dailymotion/vast-client-js/blob/master/src/parser.coffee#L32-L33)
```
@_parse url, null, options, (err, response) ->
    cb(response, err)
```

michaelvial:
Thanks for spotting this typo @Knotschi

eliamaino-fp:
Hi @Knotschi, we corrected the documentation in https://github.com/dailymotion/vast-client-js/pull/197. About the twisted order of the callback parameters in `DMVAST.parser.parse`, we will try to standardize the pattern in the next release.

PR #197, opened 2018-02-22
callback of DMVAST.parser.load is called with params in wrong order

eliamaino-fp:
See issue https://github.com/dailymotion/vast-client-js/issues/193


michaelvial:
👍

PR #201, opened 2018-02-28
Use webpack for bundle

eliamaino-fp:
This PR introduces webpack as tool to bundle. This will allow a more robust bundle process and will make easier to progressively decaffeinate the project.


src/urlhandler.coffee
```diff
@@ -18,7 +18,7 @@ class URLHandler
             return options.urlhandler.get(url, options, cb)
         else if not window?
             # prevents browserify from including this file
-            return require('./urlhandlers/' + 'node').get(url, options, cb)
+            return require('./urlhandlers/' + 'node.coffee').get(url, options, cb)
```
rumesh: Put in one string

Issue #204, opened 2018-03-16
Persistent storage handling


mseeley:
Hello Dailymotion 👋 . Our application imports the vast-client lib. Thank you very much for sharing it.

I have several observations I would appreciate your feedback on.

1. `VASTUtil.storage` requires `localStorage` or `sessionStorage`. We run on embedded devices. It's not unusual for devices to have experimental `localStorage` implementations. Have you considered allowing injection of different storage implementation that match the existing interface?
1. The keys for values stored by `vast-client` are not partitioned away from other keys. Eg, `lastSuccessfullAd`, `totalCalls`, `totalCallsTimeout` can overwrite other environment values or be themselves be overwritten unintentionally. This is a good opportunity for the `vast-client` to prefix its keys.
1. `__VASTUtil__` the key/value is set and retrieved but never removed. Desktop browsers have gobs of storage while embedded devices have far far less. While this is a little key and little value it would be kind to the host environment and app to cleanup after testing for a storage interface.

Thanks for your feedback.

eliamaino-fp:
Hey @mseeley thanks for these suggestions. I'll go in order:

1.  We never had to deal with embedded `localStorage` implementations for the moment. I agree that it could be a good idea to provide a `storage` interface to allow to inject a custom implementation.
2. Yes, the current implementation is not the safest in terms of naming. We'll take it in consideration
3. Totally agree :)

Anyway we're currently working on a `2.0` release which will introduce many changes in the current APIs, therefore thanks for the feedback we'll definitely take it into account in the process.

eliamaino-fp:
Custom storage support is now available from `2.0` release, see https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-client.md#storage-storage

PR #206, opened 2018-03-26
VASTClient API improvements

eliamaino-fp:
This PR changes the internal logic and the API of VASTClient

Has to be merged in the feature branch `api-improvements` (https://github.com/dailymotion/vast-client-js/tree/api-improvements)

Main changes

- [x]  Set `VASTClient` class properties through constructor
- [x]  Use error first callbacks
- [x]  Provide support for custom storage (https://github.com/dailymotion/vast-client-js/issues/204):
Creation of a `Storage` class separated from utils. Moved all utility classes under `util` folder.
- [x]  Improve documentation using JsDoc comments (only for `VASTClient` and `Storage` classes)
- [x]  Added and improved tests. Test file renamed to `test/vast_client.js`
- [x]  File renamed to `vast_client.js` in order to match class name


src/util/util.js
```diff
@@ -1,69 +1,4 @@
...
-        typeof window !== 'undefined' && window !== null
-          ? window.localStorage || window.sessionStorage
-          : null;
-    } catch (storageError) {
-      storage = null;
-    }
-
-    // In Safari (Mac + iOS) when private browsing is ON,
-    // localStorage is read only
-    // http://spin.atomicobject.com/2013/01/23/ios-private-browsing-localstorage/
```
rumesh: Keep this comment, it can be usefull

src/vast_client.js
```diff
@@ -0,0 +1,130 @@
...
+    this.cappingMinimumTimeInterval = cappingMinimumTimeInterval || 0;
+    this.defaultOptions = {
+      withCredentials: false,
+      timeout: 0
+    };
+    this.vastParser = new VASTParser();
+    this.util = new Util();
+    this.storage = customStorage || new Storage();
+
+    // Init values if not already set
```
rumesh: How can it be already set ? we are in the constructor

src/vast_client.js
```diff
@@ -0,0 +1,130 @@
...
+        cb = options;
+        options = {};
+      } else {
+        throw new Error(
+          'VASTClient get method called without valid callback function'
+        );
+      }
+    }
+
+    const now = +new Date();
```
rumesh: Use Date.now(), it's better than instantiate new Date

src/vast_client.js
eliamaino-fp: Yes we are, but these three properties are retrieved from the `storage` through getter methods. Therefore it can be that we already have values for them at this point of execution.

PR #213, opened 2018-04-18
Quartile Fix

rumesh:
If ads seeked (IOS fullscreen) be sure to trigger all quartile regarding progress time

### DIFF
--- a/src/vast_tracker.js
+++ b/src/vast_tracker.js
@@ -48,6 +48,8 @@ export class VASTTracker extends EventEmitter {
       'closeLinear',
       'close'
     ];
+    // Have to save already triggered quartile, to not trigger again
+    this._alreadyTriggeredQuartiles = {};
     // Duplicate the creative's trackingEvents property so we can alter it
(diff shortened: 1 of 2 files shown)


src/vast_tracker.js
```diff
@@ -156,6 +160,24 @@ export class VASTTracker extends EventEmitter {
     this.progress = progress;
   }
 
+  /**
+   * Check if a quertile is reached, and not already triggered
```
eliamaino-fp: "Checks if a quartile has been reached without have being triggered already."

src/vast_tracker.js
```diff
@@ -156,6 +160,24 @@ export class VASTTracker extends EventEmitter {
...
+   *
+   * @param {String} quartile - Quartile name
+   * @param {Number} time - Time offset, when this quartile is reached
+   * @param {Number} progress - Current progress of the ads
```
eliamaino-fp: For `time` and `progress` parameters, let's specify if the given time is in seconds or milliseconds

PR #217, opened 2018-05-09
Better conditional checks

eliamaino-fp:
After decaiffeinating the project many conditional checks bacame too verbose. This PR makes them more readable

### DIFF
--- a/src/parser/ad_parser.js
+++ b/src/parser/ad_parser.js
@@ -175,7 +175,7 @@ export class AdParser {
       'VASTAdTagURI'
     );
 
-    if (wrapperURLElement != null) {
+    if (wrapperURLElement) {
       ad.nextWrapperURL = this.parserUtils.parseNodeText(wrapperURLElement);
     } else {
       wrapperURLElement = this.parserUtils.childByName(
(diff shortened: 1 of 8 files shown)


src/parser/ad_parser.js
```diff
@@ -175,15 +175,15 @@ export class AdParser {
...
+    if (wrapperURLElement) {
       ad.nextWrapperURL = this.parserUtils.parseNodeText(wrapperURLElement);
     } else {
       wrapperURLElement = this.parserUtils.childByName(
         wrapperElement,
         'VASTAdTagURL'
       );
 
-      if (wrapperURLElement != null) {
+      if (wrapperURLElement) {
```
briganti: 👋 
Why do we do:
- `ad.nextWrapperURL = this.parserUtils.parseNodeText(wrapperURLElement);` line 179
and
- `ad.nextWrapperURL = this.parserUtils.parseNodeText(this.parserUtils.childByName(wrapperURLElement, 'URL'));` lines 187-189

Shouldn't we give a different name to `wrapperURLElement` line 181 cause it seems it's not the same things as the one line 178.

Bisous du canada

eliamaino-fp: 👋 
Hola Micha!

I agree on this, actually this is still the logic we had before in the coffeescript implementation https://github.com/dailymotion/vast-client-js/blob/master/src/parser.coffee#L219

I'll take a look :)

Bisous de Paris

PR #220, opened 2018-05-17
Back port work from @walkerflocker to 2.x branch

rumesh:
Original commit from master https://github.com/dailymotion/vast-client-js/commit/6d3e2b547f262995b44a5e51c2e0b2c2fbf34e1c


src/parser/vast_parser.js
```diff
@@ -185,6 +185,7 @@ export class VASTParser extends EventEmitter {
     const vastResponse = new VASTResponse();
     const childNodes = vastXml.documentElement.childNodes;
 
+    let countParsedAds = 0;
```
eliamaino-fp: why not `const`?

src/parser/vast_parser.js
```diff
@@ -209,6 +211,20 @@ export class VASTParser extends EventEmitter {
       }
     }
 
+    const vastAdsLength = vastResponse.ads.length;
+    const lastAddedAds = vastResponse.ads[vastAdsLength - 1];
```
eliamaino-fp: what do you think about:
```Javascript
const lastAddedAdIndex = vastResponse.ads.length - 1;
const lastAddedAd = vastResponse.ads[lastAddedAdIndex];
```

Anyway no big opinion, choose the one you prefer :p

src/parser/vast_parser.js
```diff
@@ -209,6 +211,20 @@ export class VASTParser extends EventEmitter {
...
+    // if in child nodes we have only one ads
+    // and wrapperSequence is defined
+    // and this ads doesn't already have sequence
+    if (
+      countParsedAds === 1 &&
+      options.wrapperSequence !== undefined &&
+      options.wrapperSequence !== null &&
+      !lastAddedAds.sequence
```
eliamaino-fp: we need to check for `lastAddedAds` existance before, since it can be `undefined` we may have `Uncaught TypeError: Cannot read property 'sequence' of undefined`

So just need to add 
```Javascript
lastAddedAds &&
!lastAddedAds.sequence
```

test/vast_parser.js
```diff
@@ -782,6 +784,41 @@ describe('VASTParser', function() {
...
+      before(done => {
+        vastParser.addURLTemplateFilter(url => {
+          this.templateFilterCalls.push(url);
+          return url;
+        });
+        vastParser.getAndParse(
+          urlfor('wrapper-sequence.xml'),
+          (err, response) => {
+            if (err) {
+              console.log(err);
```
eliamaino-fp: no need for this log

test/vast_parser.js
```diff
@@ -782,6 +784,41 @@ describe('VASTParser', function() {
...
     });
 
+    describe('#For the wrapper-1 ad', function() {
+      this.response = null;
+      this.templateFilterCalls = [];
+
+      before(done => {
+        vastParser.addURLTemplateFilter(url => {
+          this.templateFilterCalls.push(url);
+          return url;
```
eliamaino-fp: I think we can avoid this line

src/parser/vast_parser.js
eliamaino-fp: I am dumb

test/vast_parser.js
rumesh: NO 👼

PR #221, opened 2018-06-04
Use promises to handle wrapper resolving

eliamaino-fp:
This PR improves the wrapper resolving logic using `Promises` instead of passing the `callback` function in the recursive calls.

This allows to:
- Resolve multiple wrappers
- Correctly support ad Pods

**Tasks**
- [x] Update JSDOC
- [x] Update documentation
- [x] Add some tests


src/parser/vast_parser.js
```diff
@@ -28,6 +28,10 @@ export class VASTParser extends EventEmitter {
     this.maxWrapperDepth = null;
     this.URLTemplateFilters = [];
     this.parentURLs = [];
+    this.parentURLs = [];
```
TheAddictedOne: Duplicate line?

src/parser/vast_parser.js
```diff
@@ -89,12 +93,11 @@ export class VASTParser extends EventEmitter {
    * Fetches a VAST document for the given url.
    * Returns a Promise which resolves,rejects according to the result of the request.
    * @param  {String} url - The url to request the VAST document.
-   * @param  {Object} options - An optional Object of parameters to be used in the request.
    * @emits  VASTParserr#VAST-resolving
```
TheAddictedOne: `VASTParserr` with 2 `r` everywhere. Is it a typo?

src/parser/vast_parser.js
```diff
@@ -125,254 +142,279 @@ export class VASTParser extends EventEmitter {
    * @emits    VASTParserr#VAST-resolved
    * @callback cb
    */
-  getAndParse(url, options, cb) {
+  getAndParseVAST(url, options, cb) {
     if (!cb) {
```
TheAddictedOne: It means its a function with 2 or 3 params, the last one being the callback?  
I'm not a big fan of this pattern. Though, I haven't thought of a better way to handle it.  
It has nothing to do with the PR, so not a big deal.

src/parser/vast_parser.js
```diff
@@ -125,254 +142,279 @@ export class VASTParser extends EventEmitter {
...
         options.originalUrl = url;
-        this.parse(xml, options, cb);
+        this.parse(xml, options)
+          .then(ads => {
+            const response = new VASTResponse();
+            response.ads = ads;
+            response.errorURLTemplates = this.errorURLTemplates;
+            this.completeWrapperResolving(response);
+
+            cb(null, response);
```
TheAddictedOne: `Node.js` style? ('👅')

src/parser/vast_parser.js
```diff
@@ -125,254 +142,279 @@ export class VASTParser extends EventEmitter {
...
    * Executes the callback with either an error or the fully parsed VASTResponse.
-   * @param    {Object} vastXml - An object representing an xml document.
+   * @param    {Object} vastXml - An object representing a vast xml document.
    * @param    {Object} options - An optional Object of parameters to be used in the parsing process.
    * @emits    VASTParserr#VAST-resolving
    * @emits    VASTParserr#VAST-resolved
    * @callback cb
    */
-  parse(vastXml, options, cb) {
+  parseVAST(vastXml, options, cb) {
```
TheAddictedOne: Looks like slightly different of `getAndParseVAST()` because it does not fetch the URL. But where is it used?

src/parser/vast_parser.js
```diff
@@ -125,254 +142,279 @@ export class VASTParser extends EventEmitter {
...
-  resolveWrappers(vastResponse, options, cb) {
-    const wrapperDepth = options.wrapperDepth++;
-
-    // Resolve all the wrappers
-    for (let i = 0; i < vastResponse.ads.length; i++) {
-      const ad = vastResponse.ads[i];
+  resolveWrappers(ad, wrapperDepth, originalUrl) {
+    return new Promise((resolve, reject) => {
+      // Going one level deeper in the wrapper chain
+      wrapperDepth++;
```
TheAddictedOne: I might have missed something, but `wrapperDepth` is a `Number` right?  
Which means it will always be a `cloned` of `0` if we call `parse(vastXML, { wrapperDepth: 0 })` (e.g: it will aways be equal to `1` and `maxDepth` will never be reached).  
In the previous implementation, the whole object `options` was passed to the function, which means we have a reference of it and `options.wrapperDepth` is properly incremented each time `resolveWrappers` is called?

src/parser/vast_parser.js
eliamaino-fp: It is a typo 😅

src/parser/vast_parser.js
eliamaino-fp: Yes I don't like it either. I just kept the pattern we had before. One option could be return a `Promise` which resolves with a `VASTResponse` or rejects with an `Error`.

src/parser/vast_parser.js
eliamaino-fp: `getAndParseVAST` and `parseVAST` are the 2 methods we were already exposing as public API of the parser. I just kept it in order to provide an option to parse an already fetched VAST.

src/parser/vast_parser.js
eliamaino-fp: Not really IMHO. 
- `resolveWrappers` receives `wrapperDepth` as a number and increments it. 
- Then, if if needed, it fetches another Ad and calls `parse` passing the Ad and the incremented `wrapperDepth` value. https://github.com/dailymotion/vast-client-js/pull/221/files#diff-207d2ce8628dea4d7e8aca97ef845cb8R337
-  If `parse` will need to call `resolveWrappers` again, it will call it with the incremented `wrapperDepth` value 

Mainly I am not sure I understood what you mean here: "Which means it will always be a cloned of 0 if we call parse(vastXML, { wrapperDepth: 0 })" :)

TheAddictedOne: Okay, just double-checked and I misread where the recursivity is done 😄 .  
What I meant:
```javascript
var o = { inc: 0 }
function withObject(obj) {
  obj.inc++
}
withObject(o)
o // Outputs "{ inc: 1 }" because "obj" is a reference of "o" and its value is modified

var i = 0
function withNumber(num) {
  num++
}
withNumber(i)
i // Outputs "0" because num is a copy (clone) of "i" and the initial variable is not modified
```

PR #226, opened 2018-06-29
Ad / Ad pod based wrapper resolving

eliamaino-fp:
https://github.com/dailymotion/vast-client-js/pull/221 improved our parsing algorithm allowing the client to parse all the ads contained in a VAST response in the right way.

Although this approach is correct, it may create some overhead since most of the time we just need the first Ad (or AdPod).

This PR aims to provide a way in the VASTClient component to only parse and resolve the first Ad or AdPod and provide a way to request the parsing/resolving of the following ones.

- [x]  Add a way to parse/resolve only the first Ad/AdPod in `VASTParser`
- [x]  Add a way to parse/resolve only the first Ad/AdPod in `VASTClient`
- [x]  Update documentation
- [x]  Add tests

**New API methods**
- `VASTClient.hasRemainingAds`: Returns a Boolean indicating if there are left ads to resolve for the VAST being parsed
- `VASTClient.getNextAds(all)`: Resolves and returns the following ads as a VASTResponse


src/parser/vast_parser.js
```diff
@@ -133,6 +135,31 @@ export class VASTParser extends EventEmitter {
...
+  }
+
+  /**
+   * Resolves the next group of ads. If all is true resolves all the remaining ads.
+   * @param  {Boolean} all - If true all the remaining ads are resolved
+   * @return {Promise}
+   */
+  getNextAds(all) {
+    return new Promise((resolve, reject) => {
+      if (!this.remainingAds) {
```
rumesh: this.remainingAds.length === 0

src/parser/vast_parser.js
```diff
@@ -263,9 +295,74 @@ export class VASTParser extends EventEmitter {
...
       }
 
+      this.resolveAds(ads, { resolveAll, wrapperDepth, originalUrl })
+        .then(res => resolve(res))
+        .catch(err => reject(err));
+    });
+  }
+
+  /**
+   * Resolves an Array of ads, recursively calling itself with the reamining ads if a no ad
```
rumesh: remaining ads ? typo ?

src/parser/vast_parser.js
```diff
@@ -263,9 +295,74 @@ export class VASTParser extends EventEmitter {
...
+      let adsToParse = [];
+
+      // Isolate the first Ad or AdPod and resolve only this
+      if (!resolveAll) {
+        let adPodSequenceCount = 0;
+        let isAdPodOver = false;
+
+        // We divide ads by AdsToParse and RemainingAds
+        // AdsToParse is an array either containing a single Ad or an AdPod
+        ads.forEach(ad => {
```
rumesh: Do this split only once

src/parser/vast_parser.js
vvanghelue: Is this a valid algorithm to split sequences :

```javascript
var ads = [
	{
		id: 1
	},
	{
		id: 2,
		sequence: 1
	},
	{
		id: 3,
		sequence: 2
	},
	{
		id: 4
	},
	{
		id: 5,
		sequence: 1
	},
	{
		id: 6,
		sequence: 2
	},
	{
		id: 7,
		sequence: 3
	},
	{
		id: 8,
		sequence: 1
	},
	{
		id: 9,
		sequence: 2
	},
	{
		id: 10
	},
	{
		id: 11,
		sequence: 1
	},
	{
		id: 12
	}
]

var pods = []

ads.forEach((ad) => {
  if (ad.sequence && ad.sequence > 1) {
    pods[pods.length - 1].push(ad)
    return
  }
  pods.push([ad])
})

console.log(pods)
```

test/parser_utils.js
```diff
@@ -0,0 +1,169 @@
+import path from 'path';
```
eliamaino-fp: no need for this

vvanghelue: ok i remove it

src/parser/parser_utils.js
```diff
@@ -154,48 +154,25 @@ export class ParserUtils {
...
-            adPodSequence = ad.sequence;
-          } else {
-            // If the current ad is not the next in the AdPod and not the first of an AdPod,
-            // push the previous AdPod and the current ad to the splittedVAST and clean up
-            splittedVAST.push(adPod);
-            splittedVAST.push([ad]);
-            isAdPod = false;
-            adPodSequence = null;
-            adPod = [];
+    const pods = [];
```
eliamaino-fp: we can call this `splittedVAST` since semantically it's an array of ad and adPods

src/parser/parser_utils.js
```diff
@@ -154,48 +154,25 @@ export class ParserUtils {
...
+
+    ads.forEach((ad, i) => {
+      if (ad.sequence && ad.sequence > 1) {
```
eliamaino-fp: we can simplify the check to `ad.sequence > 1`. And add a comment saying that we enter the `if` block only if we encounter an Ad which may be part of an already started AdPod

src/parser/parser_utils.js
```diff
@@ -154,48 +154,25 @@ export class ParserUtils {
...
-          isAdPod = false;
-          adPodSequence = null;
-          adPod = [];
         }
       }
+      pods.push([ad]);
     });
 
-    return splittedVAST;
+    return pods;
```
eliamaino-fp: To sum up, we can maybe make the algorithm more readable like this:
```javascript
splitVAST(ads) {
  const splittedVAST = [];
  let lastAdPod = null;

  ads.forEach((ad, i) => {
    // The current Ad may be the next Ad of an AdPod
    if (ad.sequence > 1) {
      const lastAd = ads[i - 1];
      // check if the current Ad is exactly the next one in the AdPod
      if (lastAd && lastAd.sequence === ad.sequence - 1) {
        lastAdPod && lastAdPod.push(ad);
        return;
      }
      // If the ad had a sequence attribute but it was not part of a correctly formed
      // AdPod, let's remove the sequence attribute
      delete ad.sequence;
    }
    splittedVAST.push([ad]);
    lastAdPod = splittedVAST[splittedVAST.length - 1];
  }

  return splittedVAST;
}
```

What do you think @rumesh , @vvanghelue ?

vvanghelue: 👍 far more readable, and deleting unexpected ad.sequence is a good idea too

src/parser/parser_utils.js
```diff
@@ -146,4 +146,38 @@ export class ParserUtils {
...
+   * @param  {Array} ads - An Array of ads to split
+   * @return {Array}
+   */
+  splitVAST(ads) {
+    const splittedVAST = [];
+    let lastAdPod = null;
+
+    ads.forEach((ad, i) => {
+      if (ad.sequence) {
+        ad.sequence = parseInt(ad.sequence);
```
rumesh: Always use the 2nd parameter of parseInt to specify the base, here 10 `ad.sequence = parseInt(ad.sequence, 10);`

src/parser/parser_utils.js
```diff
@@ -146,4 +146,38 @@ export class ParserUtils {
...
+        if (lastAd && lastAd.sequence === ad.sequence - 1) {
+          lastAdPod && lastAdPod.push(ad);
+          return;
+        }
+        // If the ad had a sequence attribute but it was not part of a correctly formed
+        // AdPod, let's remove the sequence attribute
+        delete ad.sequence;
+      }
+
+      splittedVAST.push([ad]);
```
rumesh: `lastAdPod = [ad];
splittedVAST.push(lastAdPod);`

better visibility of code

src/vast_client.js
```diff
@@ -68,6 +68,23 @@ export class VASTClient {
     this.storage.setItem('vast-client-total-calls-timeout', value);
   }
 
+  /**
+   * Returns a boolean indicating if there are more ads to resolve for the current parsing.
+   * @return {Boolean}
+   */
+  hasRemainingAds() {
+    return !(this.vastParser.remainingAds.length === 0);
```
rumesh: why not `this.vastParser.remainingAds.length > 0;` ?

src/parser/vast_parser.js
```diff
@@ -263,6 +319,29 @@ export class VASTParser extends EventEmitter {
         lastAddedAd.sequence = wrapperSequence;
       }
 
+      // Split the VAST in case we don't want to resolve everything at the first time
+      if (resolveAll === false) {
+        this.splittedVAST = this.parserUtils.splitVAST(ads);
+        this.remainingAds = Object.assign([], this.splittedVAST);
```
rumesh: Should be `[].concat(this.splittedVAST);` no ? as it's an array

Issue #236, opened 2018-07-23
package.json in 2.0.0


jakub-g:
In 2.0.0 `main` field in `package.json` points to `src/index.js`
https://github.com/dailymotion/vast-client-js/blob/master/package.json#L25
...which is an ES module
https://github.com/dailymotion/vast-client-js/blob/master/src/index.js
...which means AFAIU that node versions that do not support ES modules (seems like no version of node supports it now, except 10.x behind the flag?) will fail to load it.
(in 1.x `main` was pointing to `dist/index.js`).

AFAIU `main` should point to a ES5 output, whereas a new field `module` could point to the ES6 code
https://webpack.js.org/guides/author-libraries/#final-steps

Also `"browser": "vast-client.js"` should be added as it was done in 1.x branch, otherwise problems described in #156 will come back

eliamaino-fp:
I agree, so you suggest to have something like this?
```javascript
...
"main" : "vast-client.js",
"module" : "src/index.js",
"browser" : "vast-client.js"
...
```

jakub-g:
I think this should be fine:
```...
"module" : "src/index.js",
"browser" : "vast-client.js"
...```
but this
```
"main" : "vast-client.js",
```
...I guess it wouldn't work in node because it's just building the lib and exporting as a variable (check the generated `vast-client.js`) , and we want `module.exports`.

Probably we will need two separate webpack configs, for browser output and node output. I'm not that familiar how this stuff is handled typically, to be researched. Anyway, AFAIU, we'll probably end up with 3 different entry points:

```
"main" : "vast-client-node.js",
"module" : "src/index.js",
"browser" : "vast-client.js"
```

eliamaino-fp:
Fixed in https://github.com/dailymotion/vast-client-js/pull/237
And released in [`v2.0.1`](https://github.com/dailymotion/vast-client-js/releases/tag/2.0.1)

PR #237, opened 2018-07-23
Differentiate entrypoints

eliamaino-fp:
See issue https://github.com/dailymotion/vast-client-js/issues/236

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,9 @@
   "engines": {
     "node": "8.9.4"
   },
-  "main": "src/index.js",
+  "main" : "vast-client.js",
+  "module" : "src/index.js",
+  "browser" : "vast-client.js",
   "scripts": {
     "test": "mocha --compilers js:babel-core/register --reporter spec test/*.js",
     "build": "webpack",


package.json
```diff
@@ -22,10 +22,12 @@
   "engines": {
     "node": "8.9.4"
   },
-  "main": "src/index.js",
+  "main": "vast-client-node.js",
```
jakub-g: This file is not generated in the repo, you forgot to `npm run build` and check it in :)

jakub-g:
In general this seemed to be good, but I checked closely and the package size has exploded:

https://github.com/dailymotion/vast-client-js/blob/master/vast-client.js is 68 kb *unminified*
https://github.com/dailymotion/vast-client-js/blob/a953e068be2c58f080aeb02ade43aa8f68e2d615/vast-client.js is 153 kb *minified*

Then I did `$ webpack --json > stats.json` and uploaded to https://chrisbateman.github.io/webpack-visualizer/ that 72% of size are stuff from `node_modules` which come from node_url_handler and magically made their way into the bundle (check with http://webpack.github.io/analyse/)

Basically we need to get rid of node stuff from browser bundle

package.json
eliamaino-fp: We don't commit the bundled versions, only for the release https://github.com/dailymotion/vast-client-js/blob/master/docs/CONTRIBUTING.md#pull-requests. We can definitely make an exception here :p

package.json
```diff
@@ -22,10 +22,12 @@
...
+  "module": "src/index.js",
+  "browser": "vast-client.js",
```
jakub-g: Since we're changing this, maybe we should also rename those files as follows, in webpack build conf and package.json and in README / migration guide.

```
  "main": "dist/vast-client-node.js",
  "module": "src/index.js",
  "browser": "dist/vast-client-browser.js",
```

i.e. 1) move to `dist` and 2) optionally also append `-browser` in browser version. 
For people using automatic tools and `require('vast-client')`, this would non-breaking. This is breaking only for users who reference directly the filename. Anyway 2.0.0 is just out and probably no one is using it yet, so it's a good time to change if we want.

WDYT?

package.json
jakub-g: ah fine 👍 let's do it before 2.0.1 ;)

package.json
eliamaino-fp: Yes I agree with having a `dist` directory for the bundled files. And it should not be really considered a breaking change since the library is served as an `npm` package first. 👍

eliamaino-fp:
After the splitting into multiple outputs (`dist` folder):

-  **Browser minified version**: `vast-client.min.js`, 
             -  48.5kb (total size)
             - 92.9% `src` / 7.1% `node_modules`
-  **Browser un-minified version**: `vast-client.js`, 
             -  132.3kb (total size)
             - 92.9% `src` / 7.1% `node_modules`
-  **Browser minified version**: `vast-client-node.min.js`, 
             -  71.0kb (total size)
             - 64.2% `src` / 35.7% `node_modules`
-  **Browser un-minified version**: `vast-client-node.js`, 
             -  195.4kb (total size)
             - 64.2% `src` / 35.7% `node_modules`

eliamaino-fp:
**Browser builds:**
screen shot 2018-07-25 at 14 32 05

**Node builds:**
screen shot 2018-07-25 at 14 32 20

src/urlhandlers/mock_node_url_handler.js
```diff
@@ -0,0 +1,7 @@
+export class NodeURLHandler {
```
jakub-g: I'd add a remark in `url_handler` above this line

     import { XHRURLHandler } from './urlhandlers/xhr_url_handler';

telling that for a browser build, this will be replaced to a mock at build-time, so that the people reading the code will understand it

webpack.config.js
```diff
@@ -1,29 +1,54 @@
...
+  } else {
+    // Options for browser bundling
+    config.resolve.alias['./urlhandlers/node_url_handler'] =
+      './urlhandlers/mock_node_url_handler';
+    config.output.filename =
+      env === 'dev' ? 'vast-client.js' : 'vast-client.min.js';
+    config.output.library = 'VAST';
+    config.node = {
+      fs: 'empty'
+    };
```
jakub-g: the last 3 lines are _probably_ not needed anymore due to the mocking that was put in place

Issue #241, opened 2018-07-31
Not a constructor


S4n60w3n:
Version 2.0.1 from [npm](https://www.npmjs.com/package/vast-client). 

 ```
import { VASTParser } from 'vast-client'

const vastParser = new VASTParser();
```

From VastParser [documentation](https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-parser.md#example).

Thows an error

> VastParser is not a constructor

Same with VastClient...

Can you please make sure that it works for you, from clean installation?

eliamaino-fp:
Hi @S4n60w3n, can you provide us some more info about your environment and how you import the library?

Are you doing `import { VASTParser } from 'vast-client'`?

S4n60w3n:
@eliamaino-fp Yes I do, the parsing went wrong.... I will try to create minimal repo

S4n60w3n:
I have prepared [minimal repo](https://github.com/S4n60w3n/vastClient/tree/master) . The issue is little bit different, but still similar enough. 
`...VASTParser is not a constructor`

eliamaino-fp:
Thanks for the repo, I'll take a look into it as soon as possible.

eliamaino-fp:
So it looks like webpack is not using the `module` entrypoint in `package.json` when `browser` and `main` entrypoints are provided too (see https://github.com/webpack/webpack/issues/4674)

According to the thread linked above this PR should fix the issue https://github.com/dailymotion/vast-client-js/pull/243

@S4n60w3n could you try to test that on your side too?

S4n60w3n:
@eliamaino-fp  Seems to be working and minimal repo. Thanks

eliamaino-fp:
Issue fixed in https://github.com/dailymotion/vast-client-js/releases/tag/v2.0.2

S4n60w3n:
@eliamaino-fp Glad, I could help! Thanks for quick issue fix!

PR #243, opened 2018-08-01
Fix main vs module vs browser usage

eliamaino-fp:
See https://github.com/dailymotion/vast-client-js/issues/241

To test that this PR fixes all the supported environments:

1. Import the prebundled version of the client and try to use it
2. Install the library using npm and try to use it in a browser environment
3. Install the library using npm and try to use it in a node environment


src/urlhandlers/xhr_url_handler.js
```diff
@@ -1,9 +1,13 @@
...
-      // check CORS support
-      return xhr;
+    try {
+      const xhr = new window.XMLHttpRequest();
+      if ('withCredentials' in xhr) {
+        // check CORS support
+        return xhr;
+      }
+    } catch (err) {
+      return false;
```
jakub-g: maybe this should `return null` and log an error? Not sure what exactly to do here but returning boolean doesn't seem great to me. In which case can we reach `catch` here?

eliamaino-fp: Yes right 👍 

We reach the `catch` when we try to do `new window.XMLHttpRequest();` in an environment where `window` is not defined

PR #244, opened 2018-08-08
Refine the code example in README

springuper:
the code examples don't work actually, this PR fixes them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@ vastClient.get('https://www.examplevast.com/vast.xml')
   })
   .catch(err => {
     // Deal with the error
-  })
-});
+  });
 ```
 In addition to fetching and parsing a VAST resource, **VASTClient** provides options to filter a sequence of calls based on count and time of execution, together with the possibility to track URLs using **VASTTracker**.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -52,8 +51,7 @@ vastParser.parseVAST(vastXml, (err, res) => {
   })
   .catch(err => {
     // Deal with the error
-  })
-});
+  });
```
eliamaino-fp: This is actually not good enough, since the method call still shows some syntax from the old callback-based interface.

So the line `48`
```javascript
vastParser.parseVAST(vastXml, (err, res) => {
```
Should become
```javascript
vastParser.parseVAST(vastXml)
```

eliamaino-fp:
Hey @springuper thanks for the PR, can you please reword the commit messages by following the guidelines? :) https://github.com/dailymotion/vast-client-js/blob/master/docs/CONTRIBUTING.md#commit-messages

README.md
springuper: @eliamaino-fp fixed, please CR again.

PR #250, opened 2018-10-19
[parser] Fix wrapper timeout/URI unavailable causing the parsing to fail

bwallberg:
If the first VAST URI in a wrapper timed-out/is unavailable the entire parsing fails
because the catch() that handles this starts in the second wrapper chain, not the first.

### DIFF
--- a/src/parser/vast_parser.js
+++ b/src/parser/vast_parser.js
@@ -412,33 +412,35 @@ export class VASTParser extends EventEmitter {
 
       this.fetchVAST(ad.nextWrapperURL)
         .then(xml => {
-          this.parse(xml, { originalUrl, wrapperSequence, wrapperDepth })
-            .then(unwrappedAds => {
-              delete ad.nextWrapperURL;
-              if (unwrappedAds.length === 0) {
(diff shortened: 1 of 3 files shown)


eliamaino-fp:
Hi @bwallberg thanks for opening this and sorry for the late reply, if I understood well you're saying that this `catch`
```javascript
.catch(err => {
  // Timeout of VAST URI provided in Wrapper element, or of VAST URI provided in a subsequent Wrapper element.
  // (URI was either unavailable or reached a timeout as defined by the video player.)
  ad.errorCode = 301;
  ad.errorMessage = err.message;
  resolve(ad);
});
```
Should be the one of `fetchVAST` call instead of the one of `parse` call.

Makes sense to me, would you mind adding an unit test for this scenario?

bwallberg:
@eliamaino-fp Certainly, I think I'll have time to do that tomorrow 👍

bwallberg:
@eliamaino-fp I've added a unit-test, it's basically the same as the test for an invalid vast, but with an unavailable URL instead - which should behave the same.

PR #255, opened 2018-11-19
Include more information in VAST events

kobawan:
- Updated "VAST-resolving" event to contain information on the original url and wrapper depth.
- Updated "VAST-resolved" event to contain information on errors.
- Updated unit tests
- Add documentation

### DIFF
--- a/docs/api/vast-parser.md
+++ b/docs/api/vast-parser.md
@@ -45,17 +45,37 @@ const vastParser = new VASTParser();
 ```
 
 ## Events<a name="events"></a>
-`VASTParser` extends [`EventEmitter`](https://nodejs.org/api/events.html#events_class_eventemitter), therefore is possible to add event listeners like this:
+`VASTParser` extends [`EventEmitter`](https://nodejs.org/api/events.html#events_class_eventemitter), therefore is possible to add event listeners.
+Here is the list of event emitted by the class:
+### VAST-error
+
+Event listener that provides an object with error information such as error code, error message, extensions and system.
+
 ```Javascript
-vastParser.on('VAST-error', () => {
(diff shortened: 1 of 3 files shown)


docs/api/vast-parser.md
```diff
@@ -45,17 +45,37 @@ const vastParser = new VASTParser();
 ```
 
 ## Events<a name="events"></a>
-`VASTParser` extends [`EventEmitter`](https://nodejs.org/api/events.html#events_class_eventemitter), therefore is possible to add event listeners like this:
+`VASTParser` extends [`EventEmitter`](https://nodejs.org/api/events.html#events_class_eventemitter), therefore is possible to add event listeners.
+Here is the list of event emitted by the class:
+### VAST-error
+
+Event listener that provides an object with error information such as error code, error message, extensions and system.
```
eliamaino-fp: To be 'strict' here we're documenting an event and not an event listener ;)

We can rephrase it to something like:

The event is triggered whenever the parser "bla bla bla (explain when the event should be fired)" and carries the following data:
- `ERRORCODE [optional]` - bla bla bla

Same for the other events

docs/api/vast-parser.md
```diff
@@ -45,17 +45,37 @@ const vastParser = new VASTParser();
...
+
+```Javascript
+vastParser.on('VAST-resolved', ({ url, error }) => {
+  // Access to the info
+});
+```
+
+### VAST-resolving
+
+Event listener that provides an object with information such as the current url of VAST, current depth of wrapper and original url of VAST.
```
eliamaino-fp: *current depth of the wrapper chain

kobawan: changed

PR #257, opened 2018-12-03
Check if resolveURL is a string.

alexpashkov:


eliamaino-fp:
Looks ok @alexpashkov. Can you just rename the commit by following [the guidelines](https://github.com/dailymotion/vast-client-js/blob/master/docs/CONTRIBUTING.md#commit-messages) please ? :)

And what about adding an unit test to check this case?

PR #262, opened 2018-12-21
Re-organize the code

eliamaino-fp:
Let's try to make some improvements on the codebase :)

- [x] Reduce verboseness of Promise-based functions
- [x] Slim some files (code splitting)
- [x] Functions over classes

#### URL Handlers
- [**`NodeURLHandler`**](https://github.com/dailymotion/vast-client-js/blob/master/src/urlhandlers/node_url_handler.js) from `class` to `functions`
- [**`MockNodeURLHandler`**](https://github.com/dailymotion/vast-client-js/blob/master/src/urlhandlers/mock_node_url_handler.js) from `class` to `functions`
- [**`XHRURLHandler`**](https://github.com/dailymotion/vast-client-js/blob/master/src/urlhandlers/xhr_url_handler.js) from `class` to `functions`
- [**`FlashURLHandler`**](https://github.com/dailymotion/vast-client-js/blob/master/src/urlhandlers/flash_url_handler.js) from `class` to `functions`
- [**`URLHandler`**](https://github.com/dailymotion/vast-client-js/blob/master/src/url_handler.js) from `class` to `functions`

#### Creative parsers
- [**`AdParser`**](https://github.com/dailymotion/vast-client-js/blob/master/parser/ad_parser.js) from `class` to `functions`
- [**`CreativeLinearParser`**](https://github.com/dailymotion/vast-client-js/blob/master/parser/creative_linear_parser.js) from `class` to `functions`
- [**`CreativeNonLinearParser`**](https://github.com/dailymotion/vast-client-js/blob/master/parser/creative_non_linear_parser.js) from `class` to `functions`
- [**`CompanionParser`**](https://github.com/dailymotion/vast-client-js/blob/master/parser/companion_parser.js) from `class` to `functions`

#### Utils
- [**`Util`**](https://github.com/dailymotion/vast-client-js/blob/master/util.js) from `class` to `functions`
- [**`ParserUtils`**](https://github.com/dailymotion/vast-client-js/blob/master/parser/parser_utils.js) from `class` to `functions`

#### Parser
- [**`VASTParser`**](https://github.com/dailymotion/vast-client-js/blob/master/parser/vast_parser.js) reduced `Promise`s complexity. From 550 lines to 451 lines.

### DIFF
--- a/src/parser/parser_utils.js
+++ b/src/parser/parser_utils.js
@@ -1,6 +1,4 @@
-import { Util } from '../util/util';
-
-const util = new Util();
+import { util } from '../util/util';
 
 /**
  * This module provides support methods to the parsing classes.
--- a/src/parser/parser_utils.js
+++ b/src/parser/parser_utils.js
@@ -180,4 +180,73 @@ export class ParserUtils {
 
     return splittedVAST;
   }
+
+  /**
+   * Merges the data between an unwrapped ad and his wrapper.
+   * @param  {Ad} unwrappedAd - The 'unwrapped' Ad.
+   * @param  {Ad} wrapper - The wrapper Ad.
+   * @return {void}
+   */
+  mergeWrapperAdData(unwrappedAd, wrapper) {
+    unwrappedAd.errorURLTemplates = wrapper.errorURLTemplates.concat(
+      unwrappedAd.errorURLTemplates
+    );
+    unwrappedAd.impressionURLTemplates = wrapper.impressionURLTemplates.concat(
+      unwrappedAd.impressionURLTemplates
+    );
+    unwrappedAd.extensions = wrapper.extensions.concat(unwrappedAd.extensions);
+
+    unwrappedAd.creatives.forEach(creative => {
+      if (wrapper.trackingEvents && wrapper.trackingEvents[creative.type]) {
+        for (let eventName in wrapper.trackingEvents[creative.type]) {
+          const urls = wrapper.trackingEvents[creative.type][eventName];
+          if (!creative.trackingEvents[eventName]) {
+            creative.trackingEvents[eventName] = [];
+          }
+          creative.trackingEvents[eventName] = creative.trackingEvents[
+            eventName
+          ].concat(urls);
+        }
+      }
+    });
+
+    if (
+      wrapper.videoClickTrackingURLTemplates &&
+      wrapper.videoClickTrackingURLTemplates.length
+    ) {
+      unwrappedAd.creatives.forEach(creative => {
+        if (creative.type === 'linear') {
+          creative.videoClickTrackingURLTemplates = creative.videoClickTrackingURLTemplates.concat(
+            wrapper.videoClickTrackingURLTemplates
+          );
+        }
+      });
+    }
+
+    if (
+      wrapper.videoCustomClickURLTemplates &&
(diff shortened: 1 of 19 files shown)


src/url_handler.js
```diff
@@ -1,38 +1,34 @@
...
-    } else if (this.flash.supported()) {
-      return this.flash.get(url, options, cb);
-    } else {
-      return cb(
-        new Error(
-          'Current context is not supported by any of the default URLHandlers. Please provide a custom URLHandler'
-        )
-      );
-    }
+  if (options.urlhandler && options.urlhandler.supported()) {
```
kobawan: is there ever a `options.urlhandler` passed to this function? From what I saw in the code it is `fetchingOptions` only. besides in vast parser, if a custom url handler is passed, it overrides this function

test/vast_parser.js
```diff
@@ -156,9 +157,7 @@ describe('VASTParser', function() {
       ].map(item =>
         (item =>
           it(`should not return NaN for \`${item}\``, () =>
-            isNaN(vastParser.parserUtils.parseDuration(item)).should.eql(
-              false
-            )))(item)
+            isNaN(parserUtils.parseDuration(item)).should.eql(false)))(item)
```
kobawan: this line is impossible to follow without indentations. isn't also one too many `item =>`?

src/parser/parser_utils.js
```diff
@@ -1,183 +1,254 @@
...
+ * @param  {Object} node - The node to use to find a match.
+ * @param  {String} name - The name to look for.
+ * @return {Object}
+ */
+function childByName(node, name) {
+  const childNodes = node.childNodes;
 
-    for (let childKey in childNodes) {
-      const child = childNodes[childKey];
+  for (let childKey in childNodes) {
```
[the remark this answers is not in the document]
eliamaino-fp: Right, `for ... in` is not the most suitable option for a `NodeList`, but `for ... of` is completely unsupported in IE (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Browser_compatibility) and this requires a quite bad polyfill performance-wise to work (see https://daverupert.com/2017/10/for-of-loops-are-bad/).

I wanted to use the quite new `NodeList.prototype.forEach()`, but still it's not supported in old browsers ([and not at all in IE](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Browser_Compatibility)). Transpiling it is less expansive than a `for ... of` loop, but still it will increase our bundle size. 

I know the `for ... in` syntax is not cool, but in this case is probably the most performance oriented one. What do you think @kobawan ? :)

src/url_handler.js
eliamaino-fp: Yes probably we can remove this check

test/vast_parser.js
eliamaino-fp: whoa yep it's totally unreadable 😅 

But let's change it in another PR ok? I'd like to rework all the unit tests we have anyway :)

src/parser/parser_utils.js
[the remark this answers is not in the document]
kobawan: ah I see, internet explorer again... `for in` seems to be the most performant solution then...

Issue #150, opened 2016-12-20
(Feature) vastTracker.click(URL) - hand over URL


flasheye:
Hi,
when working with VPAID it would be fine to call the 
**vastTracker.click()** 
with an own URL:
**vastTracker.click(url);**

Thanks for all the work :)

PR #265, opened 2019-01-09
Clickthrough fallback

clementFrancon:
These commits are related to the following issue : https://github.com/dailymotion/vast-client-js/issues/150

We choosed to use the provided url as a fallback to avoid having the clickthrough from the vast being overrided.

### DIFF
--- a/src/vast_tracker.js
+++ b/src/vast_tracker.js
@@ -306,9 +306,10 @@ export class VASTTracker extends EventEmitter {
    * It calls the tracking URLs and emits a 'clickthrough' event with the resolved
    * clickthrough URL when done.
    *
+   * @param {String} [fallbackClickThroughURL=null] - an optional clickThroughURL template that could be used as a fallback
    * @emits VASTTracker#clickthrough
    */
-  click() {
+  click(fallbackClickThroughURL = null) {
     if (
       this.clickTrackingURLTemplates &&
       this.clickTrackingURLTemplates.length
@@ -316,12 +317,16 @@ export class VASTTracker extends EventEmitter {
       this.trackURLs(this.clickTrackingURLTemplates);
     }
 
-    if (this.clickThroughURLTemplate) {
+    // Use the provided fallbackClickThroughURL as a fallback
+    const clickThroughURLTemplate =
+      this.clickThroughURLTemplate || fallbackClickThroughURL;
+
+    if (clickThroughURLTemplate) {
       const variables = this.linear
         ? { CONTENTPLAYHEAD: this.progressFormatted() }
         : {};
       const clickThroughURL = util.resolveURLTemplates(
-        [this.clickThroughURLTemplate],
+        [clickThroughURLTemplate],
         variables
       )[0];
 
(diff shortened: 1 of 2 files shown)


src/vast_tracker.js
```diff
@@ -306,22 +306,27 @@ export class VASTTracker extends EventEmitter {
    * It calls the tracking URLs and emits a 'clickthrough' event with the resolved
    * clickthrough URL when done.
    *
+   * @param {String} [fallbackClickThroughURL=null] - an optional clickThroughURL template that could be use as a fallback
```
eliamaino-fp: "could be use" -> "could be used"

src/vast_tracker.js
```diff
@@ -306,22 +306,27 @@ export class VASTTracker extends EventEmitter {
...
     if (
       this.clickTrackingURLTemplates &&
       this.clickTrackingURLTemplates.length
     ) {
       this.trackURLs(this.clickTrackingURLTemplates);
     }
 
-    if (this.clickThroughURLTemplate) {
+    // Use the provided fallbackClickThroughURL as a fallback
+    let clickThroughURLTemplate =
```
eliamaino-fp: why not `const`?

test/vast_tracker.js
```diff
@@ -424,6 +424,26 @@ describe('VASTTracker', function() {
           _eventsSent[1].should.eql('clickthrough');
         });
       });
+
+      describe('#click with fallback', () => {
+        let fallbackClickThroughURL =
```
eliamaino-fp: `const` :)

test/vast_tracker.js
```diff
@@ -424,6 +424,26 @@ describe('VASTTracker', function() {
...
+          'http://example.com/fallback-clickthrough';
+        before(done => {
+          // Init tracker with empty data (we don't need any since we'll test a fallback)
+          this.Tracker = new VASTTracker(vastClient, {}, {});
```
eliamaino-fp: we can do `const tracker = new VASTTracker(vastClient, {}, {});`. IDK why we were storing the tracker instance as a property of the test, but this is not a good idea since it may affect other unit tests (and it does, see the changes you had to make on line 496 😉)

Ideally every describe should have his own local instance of the tracker

test/vast_tracker.js
```diff
@@ -538,4 +539,84 @@ describe('VASTTracker', function() {
...
     });
   });
+
+  describe('#clickthroughs', () => {
+    const fallbackClickThroughURL = 'http://example.com/fallback-clickthrough',
+      clickThroughURL = 'http://example.com/clickthrough';
+
+    describe('#VAST clickThrough with no fallback provided', () => {
+      const eventsSent = [];
+      before(done => {
```
eliamaino-fp: No need for `done` to be passed and called, nothing async is going on :) The same for the other tests

test/vast_tracker.js
```diff
@@ -538,4 +539,84 @@ describe('VASTTracker', function() {
...
+        // Init tracker
+        const tracker = new VASTTracker(vastClient, {}, {});
+        // Mock emit
+        tracker.emit = (event, ...args) => {
+          eventsSent.push({ event, args });
+        };
+        tracker.click();
+        done();
+      });
+      it("should'nt have any event", () => {
```
eliamaino-fp: "should not have sent any event"

Issue #186, opened 2017-08-04
Extensions subtrees get concatenated/garbled


ScottyMJacobson:
The VAST 3.0 spec isn't super descriptive on how Extensions should be parsed, but it seems to me that they should be able to become their own nested XML trees. In the example below, the vast client gives me back an `extensions` array, whose first element is an `extension`, but the `name` property of this extension becomes the name of the first tag and the `value` becomes some amalgam (with newlines) of all of the inner text tags.

```
<!-- VAST document with one InLine ad -->
<VAST version="2.0">
    <Ad id="601364">
      <InLine>
        <AdSystem>Acudeo Compatible</AdSystem>
        <AdTitle>VAST 2.0 Instream Test 1</AdTitle>
        <Description>VAST 2.0 Instream Test 1</Description>
        <Creatives>
            <Creative id="6012" AdID="601364" sequence="1">
                <Linear>
                    <Duration>00:00:30</Duration>
                    <MediaFiles>
                        <MediaFile id="firstFile" delivery="progressive" type="video/x-flv" bitrate="500" width="400" height="300" scalable="true" maintainAspectRatio="true"  apiFramework="VPAID">
                          <![CDATA[http://cdnp.tremormedia.com/video/acudeo/Carrot_400x300_500kb.flv]]>
                        </MediaFile>
                    </MediaFiles>
                </Linear>
            </Creative>
        </Creatives>
        <Extensions>
            <Extension type="FoodsExtension">
                <Foods>
                    <Vegetables>
                        <Vegetable>Carrot</Vegetable>
                        <Vegetable>Beets</Vegetable>
                    </Vegetables>
                    <Fruits>
                        <Fruit>Apple</Fruit>
                        <Fruit>Strawberry</Fruit>
                    </Fruits>
                </Foods>
            </Extension>
        </Extensions>
       </InLine>
      </Ad>
</VAST>
```

michaelvial:
Yes this is something we need to rework

clementFrancon:
Hello, 
FYI, this issue have been fixed and is currently on a PR : https://github.com/dailymotion/vast-client-js/pull/267

PR #267, opened 2019-01-16
Refactor of how extentions are parsed to handle nested xml

clementFrancon:
These commits are related to the following issue : https://github.com/dailymotion/vast-client-js/issues/186

This is supposed to now handle nested elements on extensions without breaking everything already here.


src/parser/ad_parser.js
```diff
@@ -232,47 +232,62 @@ function parseWrapper(wrapperElement) {
...
+ */
+function parseExtension(extNode, isChild = false) {
+  // Ignore comments
+  if (extNode.nodeName === '#comment') return null;
+
+  let ext;
+  const extNodeAttrs = extNode.attributes;
+  const childNodes = extNode.childNodes;
+
+  // Only children have a name an values
```
kobawan: minor typo: `// Only children have a name and values`

test/vast_parser.js
```diff
@@ -1210,6 +1210,68 @@ describe('VASTParser', function() {
...
+          ad = response.ads[0];
+          done();
+        });
+    });
+
+    it('should have 2 extensions', () => {
+      ad.extensions.should.have.length(2);
+    });
+
+    it('validate first extension', () => {
```
kobawan: maybe it would also be good to check that the extension doesn't have a name or value attribute

eliamaino-fp:
Oh and please @clementFrancon add a quick description to the PR and a link to the related issue. Thanks :)

src/ad_extension.js
```diff
@@ -3,4 +3,9 @@ export class AdExtension {
     this.attributes = {};
     this.children = [];
   }
+  isEmpty() {
+    return !(
+      Object.keys(this.attributes).length > 0 || this.children.length > 0
+    );
+  }
```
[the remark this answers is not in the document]
clementFrancon: You're right, I simply copied/pasted the other one and removed not needed elements. My bad, I'll fix this

src/ad_extension.js
```diff
@@ -3,4 +3,9 @@ export class AdExtension {
     this.attributes = {};
     this.children = [];
   }
+  isEmpty() {
```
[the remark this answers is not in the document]
kobawan: sure, but no worries, types in definitelyTyped mention the version they support. I can make a note to take care of it

src/parser/ad_parser.js
```diff
@@ -232,47 +232,62 @@ function parseWrapper(wrapperElement) {
...
-
-          for (let extChildNodeAttrKey in childNodeAttributes) {
-            const extChildNodeAttr = childNodeAttributes[extChildNodeAttrKey];
+/**
+ * Parses an extension child node
+ * @param {Object} extNode - The extension node to parse
+ * @param {Boolean} isChild - A flag to tell how to parse the node
+ * @return {Object}
+ */
+function parseExtension(extNode, isChild = false) {
```
kobawan: There seems to be a problem with each iteration of a node in the extension. For each iteration, it adds an extra:
```
{
    "name": null,
    "value": null,
    "attributes": {},
    "children": [
         {
              "value": 0,
              "attributes": {},
               "children": []
         }
     ]
},
```

src/parser/ad_parser.js
```diff
@@ -232,47 +232,62 @@ function parseWrapper(wrapperElement) {
...
+  const extNodeAttrs = extNode.attributes;
+  const childNodes = extNode.childNodes;
+
+  // Only children have a name and value
+  if (isChild) {
+    ext = new AdExtensionChild();
+    const txt = parserUtils.parseNodeText(extNode);
+
+    if (txt !== '') {
+      ext.value = txt;
```
kobawan: if there are many nested nodes, like: 
```
<Extension type="FoodsExtension">
  <Foods>
    <Vegetables>
      <Vegetable>Carrot</Vegetable>
      <Vegetable>Beets</Vegetable>
    </Vegetables>
    <Fruits>
      <Fruit>Apple</Fruit>
      <Fruit>Strawberry</Fruit>
    </Fruits>
  </Foods>
</Extension>
```
Node `<Foods>` will have value: `"value": "Carrot\n                        Beets\n                    \n                    \n                        Apple\n                        Strawberry",`.
So maybe we could try to find if that node contains valuable `value`, like `#text` or `#cdata-section`.

src/parser/ad_parser.js
```diff
@@ -237,47 +236,70 @@ function parseWrapper(wrapperElement) {
  */
 function parseExtensions(collection, extensions) {
```
kobawan: I don't think we should pass collection to this function. we should just return extension, and in the place we call `parseExtensions` we check if the returned value is truthy, then push it to `ad.extensions`. This is because it isn't clear that we are mutating `ad.extensions` by calling this function

kobawan: Also the typings can be improved from `Array` to `Node[]` for `extensions` param

src/parser/ad_parser.js
```diff
@@ -237,47 +236,70 @@ function parseWrapper(wrapperElement) {
...
+    }
+  });
+}
+
+/**
+ * Parses an extension child node
+ * @param {Object} extNode - The extension node to parse
+ * @return {Object}
+ */
+function parseExtension(extNode) {
```
kobawan: extNode could have more explicit type of `Node`. And return should be `AdExtension|null`

src/parser/ad_parser.js
```diff
@@ -237,47 +236,70 @@ function parseWrapper(wrapperElement) {
...
+  // Ignore comments
+  if (extNode.nodeName === '#comment') return null;
+
+  const ext = new AdExtension();
+  const extNodeAttrs = extNode.attributes;
+  const childNodes = extNode.childNodes;
+  const txt = parserUtils.parseNodeText(extNode);
+
+  if (txt !== '') {
```
kobawan: I think we should only add value, if `extNode` doesn't have any children, otherwise it will give unreadable results

src/parser/ad_parser.js
```diff
@@ -237,47 +236,70 @@ function parseWrapper(wrapperElement) {
...
+    ext.value = txt;
+    ext.name = extNode.nodeName;
+  }
+
+  // Parse attributes
+  if (extNode.attributes) {
+    for (const extNodeAttrKey in extNodeAttrs) {
+      if (extNodeAttrs.hasOwnProperty(extNodeAttrKey)) {
```
kobawan: is this check necessary? I mean for...in wouldnt output an `extNodeAttrKey` if it wasnt a property of `extNodeAttrs`, but maybe I'm missing something

src/parser/ad_parser.js
```diff
@@ -232,47 +232,62 @@ function parseWrapper(wrapperElement) {
...
+  // Parse all children
+  for (let childNodeKey in childNodes) {
+    const parsedChild = parseExtension(childNodes[childNodeKey], true);
+    if (parsedChild) {
+      ext.children.push(parsedChild);
+    }
+  }
+
+  // Only return not empty objects to not pollute extentions
+  if (!ext.isEmpty()) return ext;
```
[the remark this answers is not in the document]
kobawan: Also I would prefer doing `return ext.isEmpty() ? null : ext`

test/vast_parser.js
```diff
@@ -258,16 +258,16 @@ describe('VASTParser', function() {
 
       it('validate third extension', () => {
         ad1.extensions[2].attributes['type'].should.eql('Count');
-        ad1.extensions[2].children.should.have.length(1);
-        ad1.extensions[2].children[0].name.should.eql('#cdata-section');
-        ad1.extensions[2].children[0].value.should.eql('4');
+        ad1.extensions[2].children.should.have.length(0);
```
kobawan: I would prefer to move all tests related to `parseExtensions()` to a new test file called `ad_parser.js`. If you want you can rebase your changes on https://github.com/dailymotion/vast-client-js/pull/279 and implement this new file as jest (file should have `*.spec.js` name ending and be in `spec` folder with the other jest tests).

test/vast_parser.js
```diff
@@ -1210,6 +1210,68 @@ describe('VASTParser', function() {
     });
   });
 
+  describe('#Nester xml extensions', () => {
+    let ad;
+    before(done => {
+      vastParser
+        .getAndParseVAST(urlfor('nested-xml-extensions.xml'), {
```
kobawan: you should only have to call `parseExtensions` and not depend on `getAndParseVAST`. see example `spec/creative_companion_parser.spec.js` in https://github.com/dailymotion/vast-client-js/pull/280

Issue #269, opened 2019-01-23
dist/vast-client.js misnamed xhr function


Paradam:
We've discovered an issue with the compiled vast-client.js file where creating an XHR object at line 1687 which refers to an undefined function "_xhr2()"

Expected Result
`var _xhr2 = xhr();`
Ads are requested.

Actual Result
undefined is not a function and no ads play or are requested.

Tested using latest 2.2.0 from npm.

eliamaino-fp:
Hi @Paradam , thanks for spotting this. A fix is on the way and will be released soon https://github.com/dailymotion/vast-client-js/pull/270

eliamaino-fp:
Fix was released in https://github.com/dailymotion/vast-client-js/releases/tag/2.2.1, should be ok now. Please @Paradam let us know if you still encounter the bug.

Paradam:
@eliamaino-fp The update is working for us now.
Thanks.

PR #270, opened 2019-01-23
Fix naming conflict in Flash and XHR Urlhandlers

eliamaino-fp:
### Description

https://github.com/dailymotion/vast-client-js/pull/262/commits/76991fb664707eedc3a6bab79d39ee45ff86d044 Introduced a naming conflict in `FlashURLHandler` and `XHRURLHandler`

### Issue

https://github.com/dailymotion/vast-client-js/issues/269

### Type
- [ ] Breaking change
- [ ] Enhancement
- [X] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/urlhandlers/flash_url_handler.js
+++ b/src/urlhandlers/flash_url_handler.js
@@ -1,9 +1,9 @@
 function xdr() {
-  let xdr;
+  let request;
   if (window.XDomainRequest) {
-    xdr = new XDomainRequest();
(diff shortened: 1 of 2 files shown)


Paradam:
I might be looking at it wrong, but line 27 `const xdr = xdr();` in _flash_url_handler.js_ looks like it needs to be updated to be `const request = xdr();` to match the lines below it.

PR #274, opened 2019-02-01
Replace NodeJS EventEmitter by custom one

eliamaino-fp:
### Description

The client exposes 2 classes (`VASTClient` and `VASTTracker`) which provide an event base interface. To do that the classes extend the **nodeJS** `EventEmitter` class. We would like to get rid of this 3rd party import by provide a custom light `EventEmitter` class

### Type
- [x] Breaking change
- [X] Enhancement
- [ ] Fix
- [ ] Documentation
- [X] Tooling

### DIFF
--- a/src/util/event_emitter.js
+++ b/src/util/event_emitter.js
@@ -13,34 +13,22 @@ export class EventEmitter {
    * @returns {EventEmitter}
    */
   on(event, handler) {
+    if (typeof handler !== 'function') {
+      throw new TypeError(
+        `The handler argument must be of type Function. Received type ${typeof handler}`
+      );
+    }
+    if (!event) {
+      throw new TypeError(
+        `The event argument must be of type String. Received type ${typeof event}`
+      );
+    }
     this._handlers.push({
       event,
       handler
     });
 
     return this;
-  }
-
-  _onceWrapper(...args) {
-    if (!this.fired) {
-      this.target.off(this.event, this.wrapFn);
-      this.fired = true;
-      Function.prototype.apply.call(this.handler, this.target, [...args]);
-    }
-  }
-
-  _onceWrap(event, handler) {
-    const state = {
-      fired: false,
-      wrapFn: undefined,
-      target: this,
-      event,
-      handler
-    };
-    const wrapped = this._onceWrapper.bind(state);
-    wrapped.handler = handler;
-    state.wrapFn = wrapped;
-    return wrapped;
   }
 
   /**
@@ -51,7 +39,7 @@ export class EventEmitter {
    * @returns {EventEmitter}
    */
   once(event, handler) {
-    return this.on(event, this._onceWrap(event, handler));
+    return this.on(event, onceWrap(this, event, handler));
   }
 
   /**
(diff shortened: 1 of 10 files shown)


test/event_emitter.js
```diff
@@ -0,0 +1,22 @@
+import { EventEmitter } from '../src/util/event_emitter';
+
+describe('EventEmitter', function() {
+  beforeEach(function() {
+    this.emitter = new EventEmitter();
```
kobawan: personally I'm against using `this` in unit tests, I would rather define the var outside of `beforeEach` and initialise it inside.

test/event_emitter.js
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+  describe('when a listener is added', function() {
```
kobawan: and test for removing listener?

eliamaino-fp: Yes tests are still missing. I couldn't  find the time to write them yet 😓 That's why I added the **wip** label to the PR

src/util/event_emitter.js
```diff
@@ -0,0 +1,140 @@
...
+   * Synchronously calls each of the handlers registered for the named event,
+   * in the order they were registered, passing the supplied arguments to each.
+   * @param {String} event
+   * @param  {any[]} args
+   * @returns {Boolean} true if the event had handlers, false otherwise.
+   */
+  emit(event, ...args) {
+    let called = false;
+    this._handlers.forEach(item => {
+      if (typeof item.handler === 'function') {
```
rumesh: What about do this when we add the handler ? it's probably better no ?

src/util/event_emitter.js
```diff
@@ -0,0 +1,140 @@
...
+  on(event, handler) {
+    this._handlers.push({
+      event,
+      handler
+    });
+
+    return this;
+  }
+
+  _onceWrapper(...args) {
```
rumesh: onceWrapper don't need to be a class method, it should be a normal function, or private function inside `_onceWrap`

src/util/event_emitter.js
```diff
@@ -0,0 +1,140 @@
...
+
+  _onceWrap(event, handler) {
+    const state = {
+      fired: false,
+      wrapFn: undefined,
+      target: this,
+      event,
+      handler
+    };
+    const wrapped = this._onceWrapper.bind(state);
```
rumesh:
```suggestion
  const onceWrapper = function(...args) {
    if (!this.fired) {
      this.target.off(this.event, this.wrapFn);
      this.fired = true;
      Function.prototype.apply.call(this.handler, this.target, [...args]);
    }
  }
  const wrapped = onceWrapper.bind(state);
```

If use of `this` can be confusing inside `onceWrapper`, It can be place outside `EventEmitter` class as helper function too

src/util/event_emitter.js
```diff
@@ -0,0 +1,140 @@
...
+    wrapped.handler = handler;
+    state.wrapFn = wrapped;
```
rumesh: Why not set this when initialising the state ?

eliamaino-fp:
The build is failing because of this dependency https://github.com/dailymotion/vast-client-js/pull/274/files#diff-32607347f8126e6534ebc7ebaec4853dL2443

I don't get why since when I pull the branch that dep is still pointing to the `4.17.10` version of lodash (it's probably due to some rebase+force push stuff 😓 )

@kobawan can you fix it?

src/util/event_emitter.js
kobawan: I added checks in `on` function and removed the check here

src/util/event_emitter.js
kobawan: I'll move these functions out

src/util/event_emitter.js
kobawan: I used the example from https://github.com/Gozala/events/blob/master/events.js . I'm also not sure why it is reassigned, I think it is because of the scope change?

kobawan: I removed it and simplified the once wrap functions

PR #275, opened 2019-02-04
Tracker events improvements

eliamaino-fp:
### Description

This PR improves documentation and payload data of VASTTracker events. 

### Type
- [ ] Breaking change
- [X] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/vast_tracker.js
+++ b/src/vast_tracker.js
@@ -422,7 +422,7 @@ export class VASTTracker extends EventEmitter {
     const isAlwaysEmitEvent = this.emitAlwaysEvents.indexOf(eventName) > -1;
 
     if (trackingURLTemplates) {
-      this.emit(eventName, trackingURLTemplates);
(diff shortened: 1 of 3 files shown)


src/vast_tracker.js
```diff
@@ -404,10 +404,10 @@ export class VASTTracker extends EventEmitter {
     const isAlwaysEmitEvent = this.emitAlwaysEvents.indexOf(eventName) > -1;
 
     if (trackingURLTemplates) {
-      this.emit(eventName, '');
```
vicjune: @eliamaino-fp Why was it an empty string here before ?

eliamaino-fp: Looks like legacy from the old 1.x API, this payload as empty string was not even documented in the public events API of the 2.x version of the client

PR #276, opened 2019-02-08
Fix options.urlHandler assertion

Reevn:
### Description

According to the documentation, you can pass a custom URL handler to the `options` parameter via `options.urlHandler`.

The code block responsible for applying this custom `urlhandler` is not doing the assertion correctly, leading to only `options.urlhandler` with lower-case `h` working.

### Issue

Untracked issue.

### Type
- [ ] Breaking change
- [ ] Enhancement
- [x] Fix
- [x] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-client.md
+++ b/docs/api/vast-client.md
@@ -93,6 +93,7 @@ By default the fully parsed `VASTResponse` contains all the Ads contained in the
     * `withCredentials: Boolean` - A boolean to enable the withCredentials options for the XHR and FLASH URLHandlers (default `false`)
     * `wrapperLimit: Number` - A number of Wrapper responses that can be received with no InLine response (default `0`)
     * `urlHandler: URLHandler` - Custom urlhandler to be used instead of the default ones [`urlhandlers`](../../src/urlhandlers)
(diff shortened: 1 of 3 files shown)


eliamaino-fp:
Hi @Reevn Thanks for spotting this. The code is actually doing what we want, but the documentation is definitely wrong.

Could you please fix that instead? the option should be documented as `options.urlhandler`, thanks :) 

Oh and don't forget -> https://github.com/dailymotion/vast-client-js/blob/master/docs/CONTRIBUTING.md#commit-messages

eliamaino-fp:
I thought about your changes, and even though the not camelcased param is what we supported until now, the camelcased one (`urlHandler`) makes more sense.

So we can keep both of them, and just consider `urlhandler` (to not break old implementations).

Can you then just add to the docs of `VASTClient` and `VASTParser` a reference to `urlhandler` too? (while specifying that `urlHandler` is the preferred param)

Reevn:
I've applied the contributing guidelines to my commits, changed the order of applying the urlHandler to prefer `urlHandler` instead of `urlhandler` and added `urlhandler` references to the docs.

I think just having `urlHandler` in the docs with no reference to `urlhandler` should be fine though, as it has been in the docs like that and the `urlhandler` case could just be handled internally to not have a breaking change.

Issue #278, opened 2019-02-13
Refactoring unit tests away from mocha. Jest VS Jasmine?


kobawan:
Current tests are in Mocha, and they require third-party libraries. The syntax is also different from Jest and Jasmine. Based on this, we should migrate to another testing framework.

Top current testing frameworks are Jest and Jasmine. Which one would be better for this project?

Pros Jest:
* Parallel test execution
* Constant support and maintenance by Facebook

Pros Jasmine:
* Big community
* Matured framework

eliamaino-fp:
We can give a try to Jest 👍

kobawan:
My personal preference is Jest because of the speed, and it supports DOM testing (which is needed for testing XML documents)

eliamaino-fp:
Support for Jest added in https://github.com/dailymotion/vast-client-js/pull/279, going to be released as part of the 3.0 version of the library.

PR #279, opened 2019-02-13
[tools] Update to babel 7 and add jest

kobawan:
### Description

Introduced Jest to start moving away from Mocha tests.
Upgraded babel to 7 since Jest version 24 only supports babel 7.

### Issue

https://github.com/dailymotion/vast-client-js/issues/278

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [x] Tooling


jest.config.js
```diff
@@ -0,0 +1,185 @@
...
+  clearMocks: true,
+
+  // Indicates whether the coverage information should be collected while executing the test
+  // collectCoverage: false,
+
+  // An array of glob patterns indicating a set of files for which coverage information should be collected
+  // collectCoverageFrom: null,
+
+  // The directory where Jest should output its coverage files
+  coverageDirectory: 'coverage'
```
eliamaino-fp: do we need this?

jest.config.js
```diff
@@ -0,0 +1,185 @@
...
+
+  // An array of file extensions your modules use
+  // moduleFileExtensions: [
+  //   "js",
+  //   "json",
+  //   "jsx",
+  //   "ts",
+  //   "tsx",
+  //   "node"
+  // ],
```
eliamaino-fp: no need to set this?

jest.config.js
```diff
@@ -0,0 +1,185 @@
...
+  // testEnvironmentOptions: {},
+
+  // Adds a location field to test results
+  // testLocationInResults: false,
+
+  // The glob patterns Jest uses to detect test files
+  // testMatch: [
+  //   "**/__tests__/**/*.[jt]s?(x)",
+  //   "**/?(*.)+(spec|test).[tj]s?(x)"
+  // ],
```
eliamaino-fp: do we need this?

jest.config.js
```diff
@@ -0,0 +1,185 @@
...
+  // resetMocks: false,
+
+  // Reset the module registry before running each individual test
+  // resetModules: false,
+
+  // A path to a custom resolver
+  // resolver: null,
+
+  // Automatically restore mock state between every test
+  // restoreMocks: false,
```
eliamaino-fp: what's the difference with the clearMocks option?

jest.config.js
kobawan: these are the default values already, we only need to uncomment them if we need to make changes

jest.config.js
kobawan: this is already working as default

jest.config.js
kobawan: clear mocks resets the mock.calls and mock.instances whereas restore mocks clears the mocks and removes the mock implementation itself, so if you mocked a return value, clear mock wont remove it and reset mock will. https://jestjs.io/docs/en/mock-function-api.html

jest.config.js
kobawan: I can comment it back

kobawan:
@eliamaino-fp sure the comments arent needed, I can just leave link to documentation

kobawan:
@eliamaino-fp  @babel/preset-env targets es2015 by default. In the docs: https://babeljs.io/docs/en/babel-preset-env
`Sidenote, if no targets are specified, @babel/preset-env will transform all ECMAScript 2015+ code by default.`

PR #280, opened 2019-02-14
VAST 3.0 Companion ads

kobawan:
### Description

Support VAST 3.0 Companion ads

#### Added properties:
CreativeCompanion:
- required: String|null

CompanionAd:
- assetWidth: Number|null
- assetHeight: Number|null
- expandedWidth: Number|null
- expandedHeight: Number|null
- apiFramework: String|null
- adSlotID: String|null
- adParameters: String|null
- xmlEncoded: Boolean|null

#### Renamed properties:
CompanionAd:
- staticResources: Array<{ url: String, creativeType: String|null }> (Renamed from staticResource)
- htmlResources: Array<String> (Renamed from htmlResource)
- iframeResources: Array<String> (Renamed from iframeResource)

#### Removed property:
CreativeCompanion:
- trackingEvents: Object
- type: String|null (It is now part of staticResources)

### Type
- [x] Breaking change
- [x] Enhancement
- [ ] Fix
- [x] Documentation
- [ ] Tooling


src/parser/creative_companion_parser.js
```diff
@@ -14,79 +14,89 @@ import { parserUtils } from './parser_utils';
...
-                }
-                companionAd.trackingEvents[eventName].push(trackingURLTemplate);
+      const trackingEventsElement = parserUtils.childByName(
+        companionResource,
+        'TrackingEvents'
+      );
+      if (trackingEventsElement) {
+        parserUtils
+          .childrenByName(trackingEventsElement, 'Tracking')
+          .forEach(trackingElement => {
```
eliamaino-fp: why are we using `forEach` here and `map` on line 21?

src/parser/creative_companion_parser.js
```diff
@@ -14,79 +14,89 @@ import { parserUtils } from './parser_utils';
...
-  parserUtils
+  creative.variations = parserUtils
     .childrenByName(creativeElement, 'Companion')
-    .forEach(companionResource => {
-      const companionAd = new CompanionAd();
-      companionAd.id = companionResource.getAttribute('id') || null;
-      companionAd.width = companionResource.getAttribute('width');
-      companionAd.height = companionResource.getAttribute('height');
-      companionAd.companionClickTrackingURLTemplates = [];
+    .map(companionResource => {
```
eliamaino-fp: Maybe we can make the scenario in which we call `childrenByName` + some array iteration faster, by making children extraction accepting a custom function to apply:
```javascript
function applyToChildrenByName(node, nodeName, fn) {
  const children = [];
  const childNodes = node.childNodes;

  for (const childKey in childNodes) {
    const child = childNodes[childKey];

    if (child.nodeName === name) {
      children.push(fn(child)); // This will save us one iteration
    }
  }
  return children;
}
```

Or
```javascript
function childrenByName(node, nodeName, fn) {
  const hasFunctionToApply = typeof fn === 'function'
  const children = [];
  const childNodes = node.childNodes;

  for (const childKey in childNodes) {
    const child = childNodes[childKey];

    if (child.nodeName === name) {
      hasFunctionToApply ? children.push(fn(child)) : children.push(child);
    }
  }
  return children;
}
```

What do you think about this idea?

kobawan: I think this will make the code less readable in my opinion. Mostly because we would be passing the entire parsing function as a callback :/ I also don't think that iterating twice in this case is a problem (if it would be a large array, maybe...). I don't have a very strong opinion on this though.

src/parser/creative_companion_parser.js
kobawan: because `companionAd.trackingEvents` is an object not an array

eliamaino-fp: Not sure I get what you mean, both times you chain to `childrenByName` which always returns an array.

src/parser/creative_companion_parser.js
```diff
@@ -14,79 +14,89 @@ import { parserUtils } from './parser_utils';
...
+      const attributes = {};
+      const companionAttrs = companionResource.attributes;
+      for (let i = 0; i < companionAttrs.length; i++) {
+        attributes[companionAttrs[i].nodeName] = companionAttrs[i].nodeValue;
+      }
+      const companionAd = new CompanionAd(attributes);
```
eliamaino-fp: I like this approach of passing at least the attributes to the constructor (and the class logic will take care of assigning them to the right properties).

Maybe we can factor this in an util method in order to use it in the future too, something like `getAttributes`:
```javascript
function getAttributes(node) {
  const attributes = {};
  const nodeAttributes = node.attributes;

  for (let i = 0; i < companionAttrs.length; i++) {
    attributes[nodeAttributes[i].nodeName] = nodeAttributes[i].nodeValue;
  }

  return attributes
}
```

src/parser/creative_companion_parser.js
```diff
@@ -14,79 +14,89 @@ import { parserUtils } from './parser_utils';
...
-        .childrenByName(companionResource, 'HTMLResource')
-        .forEach(htmlElement => {
-          companionAd.type =
-            htmlElement.getAttribute('creativeType') || 'text/html';
-          companionAd.htmlResource = parserUtils.parseNodeText(htmlElement);
-        });
+      companionAd.htmlResource =
+        parserUtils.parseNodeText(
+          parserUtils.childByName(companionResource, 'HTMLResource')
+        ) || null;
```
eliamaino-fp: According to the VAST spec 
> Each <Companion> element must specify at least one resource file that may be one of: 
> `<StaticResource>`, `<IFrameResource>` or `<HTMLResource>`

What is the approach we want to take when multiple resources are provided?

kobawan: I didn't understand from the specs whether you could have multiple `StaticResources` for example, or just multiple resources of different types. either way we were still only assigning one resource per type previously

src/parser/creative_companion_parser.js
kobawan: created function parseAttributes

src/parser/creative_companion_parser.js
kobawan: added support for multiple of the same resources

PR #283, opened 2019-04-04
Track custom code error

clementFrancon:
### Description

Now allow custom values for errorWithCode's VAST tracker method

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [x] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-tracker.md
+++ b/docs/api/vast-tracker.md
@@ -68,11 +68,13 @@ Here is the list of events emitted by the class:
 
 ## Public Methods 💚 <a name="methods"></a>
 
-### errorWithCode(errorCode)
+### errorWithCode(errorCode, isCustomCode)
 Sends a request to the URI provided by the VAST `<Error>` element. If an `[ERRORCODE]` macro is included, it will be substituted with `errorCode`.
+Pass `isCustomCode` as true in order to use any value. If false or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between 100 and 999 (see https://gist.github.com/rhumlover/5747417). Otherwise 900 will be used.
 
 #### Parameters
  * **`errorCode: String`** - Replaces `[ERRORCODE]` macro. `[ERRORCODE]` values are listed in the VAST specification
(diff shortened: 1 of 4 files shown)


docs/api/vast-tracker.md
```diff
@@ -68,8 +68,9 @@ Here is the list of events emitted by the class:
 
 ## Public Methods 💚 <a name="methods"></a>
 
-### errorWithCode(errorCode)
+### errorWithCode(errorCode, isCustomCode)
 Sends a request to the URI provided by the VAST `<Error>` element. If an `[ERRORCODE]` macro is included, it will be substituted with `errorCode`.
+Use the `isCustomCode` Boolean to avoid checking `[ERRORCODE]` during the macro remplacement and allow any value.
```
eliamaino-fp: Please add the parameter to the list of the parameters just below

eliamaino-fp: I'd say better:
> Pass `isCustomCode` as `true` in order to use any value. If `false` or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between `100` and `999` (see https://gist.github.com/rhumlover/5747417). Otherwise `900` will be used.

src/vast_tracker.js
```diff
@@ -314,9 +314,14 @@ export class VASTTracker extends EventEmitter {
    * If an [ERRORCODE] macro is included, it will be substitute with errorCode.
    *
    * @param {String} errorCode - Replaces [ERRORCODE] macro. [ERRORCODE] values are listed in the VAST specification.
+   * @param {Boolean} [isCustomError=false] - Flag to allow custom values on error code.
    */
-  errorWithCode(errorCode) {
-    this.trackURLs(this.ad.errorURLTemplates, { ERRORCODE: errorCode });
+  errorWithCode(errorCode, isCustomError = false) {
```
eliamaino-fp: in the doc you call the new param `isCustomCode ` while here `isCustomError `. 

Beside the discrepancy (which needs to be fixed), I'd go for `isCustomCode`

PR #284, opened 2019-04-25
[parser] split XML parsing from `parse` to `parseXML`

bwallberg:
### Description

Move the parsing of the VAST XML to a separate method `parseXML` to allow users to "cherry-pick" that part if they need the ads before resolving the wrappers. It also allows users to merge the unwrapped ad with the wrapped ad themselves if they need something custom from their ad provider.

### Issue

https://github.com/dailymotion/vast-client-js/issues/285

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-parser.md
+++ b/docs/api/vast-parser.md
@@ -263,11 +263,18 @@ vastParser.parseVAST(vastXml, options)
 These methods documentation is provided in order to make the parser internal logic clearer. It should not be considered as part of the class public API
 
 ### parse(vastXml, options)
-Parses the given xml Object into an array of ads. Returns a `Promise` which resolves with the array or rejects with an error according to the result of the parsing.
+Parses the given xml Object into an array of unwrapped ads. Returns a `Promise` which resolves with the array or rejects with an error according to the result of the parsing.
 
 #### Parameters
  * **`vastXml: Object`** - An object representing an xml document.
  * **`options: Object`** - An optional Object of parameters to be used in the parsing process.
+
+### parseVastXml(vastXml, options)
+Parses the given xml Object into an array of ads. Returns the array or throws an `Error` if an invalid VAST XML is provided.
+
+#### Parameters
+* **`vastXml: Object`** - An object representing an xml document.
+* **`options: Object`** - An optional Object of parameters to be used in the parsing process.
 
 ### resolveWrappers(ad, wrapperDepth, originalUrl)
 Resolves the wrappers for the given ad in a recursive way. Returns a `Promise` which resolves with the unwrapped ad or rejects with an error.
--- a/src/parser/vast_parser.js
+++ b/src/parser/vast_parser.js
@@ -228,33 +228,23 @@ export class VASTParser extends EventEmitter {
 
   /**
    * Parses the given xml Object into an array of ads
-   * Returns a Promise which resolves with the array or rejects with an error according to the result of the parsing.
+   * Returns the array or throws an `Error` if an invalid VAST XML is provided
    * @param  {Object} vastXml - An object representing an xml document.
    * @param  {Object} options - An optional Object of parameters to be used in the parsing process.
-   * @emits  VASTParser#VAST-resolving
-   * @emits  VASTParser#VAST-resolved
-   * @return {Promise}
+   * @return {Array}
+   * @throws {Error} `vastXml` must be a valid VAST XMLDocument
    */
-  parse(
-    vastXml,
-    {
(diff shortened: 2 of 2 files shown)


src/parser/vast_parser.js
```diff
@@ -231,20 +231,9 @@ export class VASTParser extends EventEmitter {
...
-    vastXml,
-    {
-      resolveAll = true,
-      wrapperSequence = null,
-      originalUrl = null,
-      wrapperDepth = 0,
-      isRootVAST = false
-    }
-  ) {
+  parseXML(vastXml, { isRootVAST = false }) {
```
eliamaino-fp: Since we removed any async operation we can make this method simply return an array of ads or throwing an Error in case the parsing fails

eliamaino-fp: And I would maybe rename this function to something more explicit like `parseVastXml`

bwallberg: @eliamaino-fp 
So, how should I use it in `parse()`? 🤔
Personally I like to have consistent returns of promises in a function so if we treat `parseVastXml` as not async we should probably try/catch it in  `parse()` and if we get an error return a rejected promise? Unless we assume parse() is called within a chain of course.

eliamaino-fp: I know that might sound not that important, but if we plan to provide `parseVastXml ` as part of the public API and the code is not asynchronous we should not return a Promise (since users may really need to have this code executing in a sync way).

Assuming that `parse` will be called only in a Promise chain may be risky too :/

I'd suggest to go for a `try/catch` inside `parse`, something like:
```javascript
parse(vastXml, /* Other params */) {
  let ads = []

  try {
    ads = this.parseXML(vastXml, { isRootVAST })
  } catch (e) {
    return Promise.reject(e)
  }

  // rest of the logic
}
```

What do you think @bwallberg?

bwallberg: @eliamaino-fp 
Fair enough :)
I've made the changes.

bwallberg:
> Since this PR exposes a new method (even without changing the existing APIs) and you pointed out in #285 that these changes may interest other users of the library, could you please update the docs too? https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-parser.md#public-methods-- :)

I've added some docs but I added it in the `private-methods` part ( for now ) since, to get full use out of it, you have to use the other private methods. Which should work as long as you lock down the version of the library, but one needs to be aware of the risks.

Of course I'll change it if you see it as public :)

src/parser/vast_parser.js
```diff
@@ -231,30 +231,19 @@ export class VASTParser extends EventEmitter {
...
-    vastXml,
-    {
-      resolveAll = true,
-      wrapperSequence = null,
-      originalUrl = null,
-      wrapperDepth = 0,
-      isRootVAST = false
-    }
-  ) {
+  parseVastXml(vastXml, { isRootVAST = false }) {
```
eliamaino-fp: Just 2 more things:
- Please add a `@throws` doc about the error too (see http://usejsdoc.org/tags-throws.html)
- The description should be changed to something like "Parses the given xml Object into an array of ads, throws an error in case an invalid VAST XML is provided"

bwallberg: 👍 done!
I hadn't even removed the description part saying "Returns promise".... shame on me. Good catch! :)

docs/api/vast-parser.md
```diff
@@ -263,12 +263,19 @@ vastParser.parseVAST(vastXml, options)
...
-Parses the given xml Object into an array of ads. Returns a `Promise` which resolves with the array or rejects with an error according to the result of the parsing.
+Parses the given xml Object into an array of unwrapped ads. Returns a `Promise` which resolves with the array or rejects with an error according to the result of the parsing.
 
 #### Parameters
  * **`vastXml: Object`** - An object representing an xml document.
  * **`options: Object`** - An optional Object of parameters to be used in the parsing process.
 
+### parseVastXml(vastXml, options)
+Parses the given xml Object into an array of ads. Returns a `Promise` which resolves with the array or rejects with an error according to the result of the parsing.
+
```
eliamaino-fp: The description needs to be updated here too :)

bwallberg: 🤦‍♂️ of course... sorry. 🙈

bwallberg: Done!

PR #287, opened 2019-05-07
Replace module entry in favor of browser

eliamaino-fp:
### Description

After moving to babel 7 we realized that providing a `module` entry pointing to the `src/index.js` of the lib may cause more issues than benefits. 

This PR keeps the same interface for the lib (nothing should change in your imports) but favors the already bundled files over the source code.

### Type
- [ ] Breaking change
- [ ] Enhancement
- [ ] Fix
- [ ] Documentation
- [X] Tooling


README.md
```diff
@@ -78,19 +78,31 @@ The API documentation is organized by components:
...
-## Support and compatibility
-The library is 100% written in JavaScript and the source code uses modern features like `modules`, `classes`, ecc... . Make sure your environment supports these features, or transpile the library when bundling your project.
-
 ### Pre-bundled versions
 We provide several pre-bundled versions of the client (see [`dist` directory](dist/))
 
-#### Browser
-A pre-bundled version of VAST Client JS is available: [`vast-client.js`](dist/vast-client.js) or [`vast-client.min.js`](dist/vast-client.min.js) [minified].
+#### Bundlers
+A version for js bundlers (like webpack or rollup) is available by default when adding the lib using a package manager (like npm or yarn): [`vast-client.js`](dist/vast-client.js) or [`vast-client.min.js`](dist/vast-client.min.js) [minified].
```
kobawan: there is no mention of the node version. Should we add it?

PR #289, opened 2019-05-14
Add companion creatives in wrapper to parsed creatives

kobawan:
### Description

Previously companion creatives in wrappers were being ignored, now they are passed to the parsed object under creatives array.

Also, companion click tracking urls from companion creatives in wrappers are passed to other companion creatives (with no duplicated urls).

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling


spec/util.spec.js
```diff
@@ -0,0 +1,131 @@
...
+const encodeRFC3986 = str => util.encodeURIComponentRFC3986(str);
+
+const encodedAssetURI = encodeRFC3986(assetURI);
+const encodedPlayhead = encodeRFC3986(playhead);
+const encodedTimestamp = encodeRFC3986(now);
+
+const resolve = (URLTemplate, variables) =>
+  util.resolveURLTemplates([URLTemplate], variables)[0];
+
+describe('util', function() {
```
eliamaino-fp: maybe we can avoid having this first describe? since we don't have a root `beforeEach` it might be nice to remove a level of indentation to all this code.

But then of course you'll not have these tests grouped under a `util` section, up to you

spec/util.spec.js
```diff
@@ -0,0 +1,131 @@
...
+const playhead = '00:12:30.212';
+const assetURI = 'http://example.com/linear-asset.mp4?foo=1&bar=_-[{bar';
+
+const encodeRFC3986 = str => util.encodeURIComponentRFC3986(str);
+
+const encodedAssetURI = encodeRFC3986(assetURI);
+const encodedPlayhead = encodeRFC3986(playhead);
+const encodedTimestamp = encodeRFC3986(now);
+
+const resolve = (URLTemplate, variables) =>
```
eliamaino-fp: maybe this should be inside the `resolveURLTemplates` block

src/util/util.js
```diff
@@ -109,12 +109,28 @@ function flatten(arr) {
   }, []);
 }
 
+/**
+ * Joins two arrays without duplicates
+ * @param {Array} arr1
+ * @param {Array} arr2
+ */
+function joinArrayUnique(arr1 = [], arr2 = []) {
+  const arr = (arr1 || []).concat(arr2 || []);
```
eliamaino-fp: So in case an user of this function passes some non falsy parameters, we are taking an empty array instead and use this for the concat. What is the benefit of that?

src/parser/parser_utils.js
```diff
@@ -202,7 +202,34 @@ function mergeWrapperAdData(unwrappedAd, wrapper) {
   );
   unwrappedAd.extensions = wrapper.extensions.concat(unwrappedAd.extensions);
 
+  const wrapperCompanions = (wrapper.creatives || []).filter(
+    creative => creative && creative.type === 'companion'
+  );
+  const wrapperCompanionClickTracking = wrapperCompanions.reduce(
+    (result, creative) => {
+      (creative.variations || []).forEach(variation => {
```
eliamaino-fp: Not sure this check really fits our safety needs, what if `creative.creations` is an object?
```javascript
({} || []).forEach(() => {}) // forEach is not a function
```

spec/util.spec.js
kobawan: I will leave it because I think it is clearer if it is under utils, besides jest has a nice command line interface, so it won't look messy

src/util/util.js
kobawan: it works as a safe guard, because sometimes we pass null or undefined values without knowing. Also another benefit is that it removes duplicates

src/parser/parser_utils.js
kobawan: variations is always an array

eliamaino-fp: then `creative.variations` will always vslidate to `true` and we don't need the safety check no?

src/util/util.js
eliamaino-fp: yep but shouldn't the safe guard be something like 
```javascript
Array.isArray(arr1) || []
```
?

src/parser/parser_utils.js
kobawan: I meant that it is never an object, but AFAIK it can be either array, null or undefined

src/util/util.js
kobawan: I made the change, can you validate it?

src/util/util.js
```diff
@@ -109,12 +109,31 @@ function flatten(arr) {
   }, []);
 }
 
+/**
+ * Joins two arrays without duplicates
+ * @param {Array} arr1
+ * @param {Array} arr2
+ */
```
eliamaino-fp: Let's add the return type too

kobawan: added

PR #290, opened 2019-06-03
Parse vast version

clementFrancon:
### Description

Allows the parser to get the version of the VAST.

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/class-reference.md
+++ b/docs/api/class-reference.md
@@ -6,6 +6,7 @@ This class represents a parsed VAST response.
 
 - `ads: Array<Ad>` [go to class](#ad)
 - `errorURLTemplates: Array<String>`
+- `version: String`
 
 ## Ad<a name="ad"></a>
 
--- a/docs/api/vast-client.md
+++ b/docs/api/vast-client.md
@@ -138,7 +138,8 @@ Using `get` method with default `options` will return a [`VASTResponse`](https:/
     ad3,
     ad4
   ],
-  errorURLTemplates
+  errorURLTemplates,
+  version
 }
 ```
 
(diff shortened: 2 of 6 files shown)


eliamaino-fp:
We could add a test for the new version field when we have wrappers. For what I understood it should return the version of the last parsed wrapper? but what if the first ad contains more than one wrapper?

I mean a VAST like this which version is it supposed to return?
```xml
<VAST version="2.0">
  <Ad><Wrapper></Wrapper></Ad> <!-- Points to a 3.0 VAST resource-->
  <Ad><Wrapper></Wrapper></Ad> <!-- Points to a 4.0 VAST resource-->
</VAST>
```

Maybe we should just keep the Version returned in the first VAST of the chain (if any), what do you think @clementFrancon ?

clementFrancon:
> We could add a test for the new version field when we have wrappers. For what I understood it should return the version of the last parsed wrapper? but what if the first ad contains more than one wrapper?
> 
> I mean a VAST like this which version is it supposed to return?
> 
> ```
> <VAST version="2.0">
>   <Ad><Wrapper></Wrapper></Ad> 
>   <Ad><Wrapper></Wrapper></Ad> 
> </VAST>
> ```
> 
> Maybe we should just keep the Version returned in the first VAST of the chain (if any), what do you think @clementFrancon ?

In this case it will probably be the last one parsed that will override the vast version (with the current implementation), it's maybe safer to only keep the version of the first vast from now. Maybe in a next step we can manage to keep the track of all versions encountered

PR #297, opened 2019-07-12
Support Ad Verification for VAST 4.1

unglud:
VAST 4 offers a designated space (<AdVerifications>) for inserting ad verification APIs, enabling a more streamlined process for executing files strictly intended for ad verification. Open Measurement (OM) is expected to be used for this purpose.

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [x] Documentation
- [ ] Tooling


src/parser/parser_utils.js
```diff
@@ -187,6 +187,32 @@ function splitVAST(ads) {
   return splittedVAST;
 }
 
+/**
+ * Parses the attributes and assign them to object
+ * @param  {Object} attributes attribute
+ * @param  {Object} verificationObject with properties which can be assigned
+ */
```
eliamaino-fp: please add a doc about the returned value too :)

docs/api/class-reference.md
```diff
@@ -139,3 +140,10 @@ This class represents a generic Creative. It's used as a parent class for more s
 - `attributes: Object`
 - `children: Array<AdExtension>`
 - `isEmpty(): Boolean` Returns true when none of these attributes have a value
+
+## AdVerification<a name="ad-verification"></a>
+- `apiFramework: String` The name of the API framework used to execute the AdVerification code
+- `browserOptional: Boolean` If *true*, this resource is optimized and able to execute in an environment without DOM and other browser built-ins (e.g. iOS' JavaScriptCore).
+- `parameters: string` Metadata about the current impression
```
kobawan: minor: should be capital `String`

src/parser/parser_utils.js
```diff
@@ -187,6 +187,32 @@ function splitVAST(ads) {
...
+          attribute.nodeName &&
+          attribute.nodeValue &&
+          verificationObject.hasOwnProperty(attribute.nodeName)
+      ) {
+        let value = attribute.nodeValue;
+
+        if (typeof verificationObject[attribute.nodeName] === 'boolean') {
+          value = parseBoolean(value);
+        }
+        verificationObject[attribute.nodeName] = value;
```
kobawan: instead of mutating verificationObject, you could clone it, make the modifications and return the new object

src/parser/parser_utils.js
```diff
@@ -187,6 +188,39 @@ function splitVAST(ads) {
   return splittedVAST;
 }
 
+/**
+ * Parses the attributes and assign them to object
+ * @param  {Object} attributes attribute
+ * @param  {Object} verificationObject with properties which can be assigned
+ * @return {AdVerification|null}
+ */
+function assignAttributes(attributes, verificationObject) {
```
kobawan: at first glance it looks like this function can be re-used, but actually it only applies to `AdVerification` class. you could either specify that this function is only for `AdVerification` or pass the class as an argument

unglud: Sorry, I'm not sure what do you mean exactly.

kobawan: If I understand correctly, this function checks that the attribute exists in the class `AdVerification`, and if so then assign the value to it. you could do:
```
function assignAttributes(attributes, verificationObject, verificationClassDefault) {
  // ...
  const ver = Object.assign(new verificationClassDefault, verificationObject)
}

assignAttributes(node.attributes, verification, new AdVerification)
```

But actually I don't quite understand the purpose of line 202, why is there a need to assign `AdVerification` to `verificationObject`?

unglud: Well it was your idea to clone it, no?
https://github.com/dailymotion/vast-client-js/pull/297#discussion_r305354981

kobawan: Yes, my point is `verificationObject` is already an instance of `AdVerification`, you don't have to assign it again to `AdVerification`. A simple: `const ver = { ...verificationObject }` would do.

Currently you have to assign the returned class instance of this function to where you call it, so perhaps it would make the most sense to re-use the function `parseAttributes`, and directly pass them on creating `new AdVerification()` like in https://github.com/dailymotion/vast-client-js/blob/3.0-version/src/parser/creative_companion_parser.js#L23 . This way you are only returning the attributes and not the class instance.
What do you think?

unglud: From the second thought, I decided to revert last change and actually mutate the object.
The reason is that I use this object twice
https://github.com/dailymotion/vast-client-js/blob/e0a75738601e9253dda97d6a1bb96c0e5297c3a2/src/parser/ad_parser.js#L329
and https://github.com/dailymotion/vast-client-js/blob/e0a75738601e9253dda97d6a1bb96c0e5297c3a2/src/parser/ad_parser.js#L339

So it easier to just mutate. It gives me a simple type check as well if I use real class.

I disagree, then let's start a discussion from this point again )

src/parser/parser_utils.js
```diff
@@ -187,6 +187,32 @@ function splitVAST(ads) {
   return splittedVAST;
 }
 
+/**
+ * Parses the attributes and assign them to object
+ * @param  {Object} attributes attribute
+ * @param  {Object} verificationObject with properties which can be assigned
+ */
+function assignAttributes(attributes, verificationObject) {
+  if (attributes) {
```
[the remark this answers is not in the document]
kobawan: unresolving this comment as it is valid again :)

src/parser/parser_utils.js
kobawan: this is ok too as it is simpler

kobawan:
@unglud Could you fix the conflicts? Once that is done I will merge the PR

unglud:
I don't see any conflicts here

src/parser/parser_utils.js
unglud: If disagree*

kobawan:
Could you pull the latest changes from 3.0-version? A pr was merged so it probably has conflicts with that. I cannot merge on my side.

unglud:
I did https://www.dropbox.com/s/htmyml8e30j4099/Screenshot%202019-08-06%2014.54.53.png?dl=0

kobawan:
This is very odd, I'm the only one who is having the PR showing conflicts. I will ask a colleague to merge the PR if I can't figure out why I have conflicts

kobawan:
Hi! The merge went incorrectly, and it merged ~40 commits. Could you please re-open another PR with only your commits? Sorry for the inconvenience

PR #299, opened 2019-08-02
Support Mezzanine for VAST 4.1

unglud:
To support advertising across video platforms that include long-form content and high-resolution screens, VAST 4 features include support for the raw, high-quality mezzanine file. The mezzanine file is very large and cannot be used for ad display, but ad-stitching services and other ad vendor use it to generate files at appropriate quality levels for the environment in which they play.

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [x] Documentation
- [ ] Tooling

### DIFF
--- a/src/mezzanine.js
+++ b/src/mezzanine.js
@@ -1,13 +1,22 @@
 export class Mezzanine {
     constructor () {
-        this.id = null;
-        this.fileURL = null;
-        this.delivery = 'progressive';
-        this.codec = null;
-        this.type = null;
-        this.width = 0;
-        this.height = 0;
-        this.fileSize = 0;
-        this.mediaType = '2D';
+      this.id = null;
+      this.fileURL = null;
+      this.delivery = null;
+      this.codec = null;
+      this.type = null;
+      this.width = 0;
+      this.height = 0;
+      this.fileSize = 0;
+      this.mediaType = '2D';
+
+      this.mandatoryFields = ['delivery', 'type', 'width', 'height']
     }
+
+  validate () {
+    return !this.mandatoryFields.some((name) => {
+      return this[name] === null || this[name] === 0;
+    });
+  }
+
 }
--- a/src/mezzanine.js
+++ b/src/mezzanine.js
@@ -9,14 +9,5 @@ export class Mezzanine {
       this.height = 0;
       this.fileSize = 0;
       this.mediaType = '2D';
-
-      this.mandatoryFields = ['delivery', 'type', 'width', 'height']
     }
-
-  validate () {
-    return !this.mandatoryFields.some((name) => {
-      return this[name] === null || this[name] === 0;
-    });
-  }
-
 }
--- a/src/mezzanine.js
+++ b/src/mezzanine.js
@@ -2,9 +2,9 @@ export class Mezzanine {
     constructor () {
         this.id = null;
         this.fileURL = null;
-        this.deliveryType = 'progressive';
+        this.delivery = 'progressive';
         this.codec = null;
-        this.mimeType = null;
+        this.type = null;
         this.width = 0;
         this.height = 0;
         this.fileSize = 0;
--- a/src/mezzanine.js
+++ b/src/mezzanine.js
@@ -1,0 +1,13 @@
+export class Mezzanine {
+    constructor () {
+        this.id = null;
+        this.fileURL = null;
+        this.deliveryType = 'progressive';
+        this.codec = null;
+        this.mimeType = null;
+        this.width = 0;
+        this.height = 0;
+        this.fileSize = 0;
+        this.mediaType = '2D';
+    }
+}
--- a/src/parser/creative_linear_parser.js
+++ b/src/parser/creative_linear_parser.js
@@ -165,17 +165,17 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
         const mezzanine = new Mezzanine();
         mezzanine.id = mezzanineElement.getAttribute('id');
         mezzanine.fileURL = parserUtils.parseNodeText(mezzanineElement);
-        mezzanine.deliveryType = mezzanineElement.getAttribute('delivery');
+        mezzanine.delivery = mezzanineElement.getAttribute('delivery');
         mezzanine.codec = mezzanineElement.getAttribute('codec');
-        mezzanine.mimeType = mezzanineElement.getAttribute('type');
+        mezzanine.type = mezzanineElement.getAttribute('type');
         mezzanine.width = parseInt(
-            mezzanineElement.getAttribute('width') || 0
+            mezzanineElement.getAttribute('width') || 0, 10
         );
         mezzanine.height = parseInt(
-            mezzanineElement.getAttribute('height') || 0
+            mezzanineElement.getAttribute('height') || 0, 10
         );
         mezzanine.fileSize = parseInt(
-            mezzanineElement.getAttribute('fileSize') || 0
+            mezzanineElement.getAttribute('fileSize') || 0, 10
         );
         mezzanine.mediaType = mezzanineElement.getAttribute('mediaType') || '2D';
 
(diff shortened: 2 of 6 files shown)


src/mezzanine.js
```diff
@@ -0,0 +1,13 @@
+export class Mezzanine {
+    constructor () {
+        this.id = null;
+        this.fileURL = null;
+        this.deliveryType = 'progressive';
+        this.codec = null;
+        this.mimeType = null;
```
rumesh: Shouldn't we call it `type` as in the spec ? and just comment it with this comments in the specs ?
Capture d’écran 2019-08-05 à 17 32 17
And for all fields shouldn't we put possible value in comment  ?

rumesh: And maybe put null on all mandatory fields and put the default values for the others ?
Just to distinct which one is mandotary and which on is optional WDYT ?

src/parser/creative_linear_parser.js
```diff
@@ -158,6 +159,28 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
...
+      const mezzanineElement = parserUtils.childByName(mediaFilesElement, 'Mezzanine');
+      if (mezzanineElement) {
+        const mezzanine = new Mezzanine();
+        mezzanine.id = mezzanineElement.getAttribute('id');
+        mezzanine.fileURL = parserUtils.parseNodeText(mezzanineElement);
+        mezzanine.deliveryType = mezzanineElement.getAttribute('delivery');
+        mezzanine.codec = mezzanineElement.getAttribute('codec');
+        mezzanine.mimeType = mezzanineElement.getAttribute('type');
+        mezzanine.width = parseInt(
+            mezzanineElement.getAttribute('width') || 0
```
rumesh: As it's a mandotory field it should raise an error if width / height is not defined

rumesh: And `parseInt` must always have the second parameters to specify decimal base, otherwise it will try to guess it and can create some bugs.

unglud: > As it's a mandotory field it should raise an error if width / height is not defined

It's ok not to crash the whole library on the minor issue. VAST is a standard, but not all agencies strictly follow the standards unfortunately.

rumesh: Yes of course it shouldn't crash but if mandatory field are not provided this mezzanine should be ignored. Because lot of agencies don't respect the standard, vast-client have to respect it. I agree to not crash everything but if someone else open a PR saying my agencies want default value at 1280/720 because our ads are all in HD what should we do ? That's why we try to be as close as to the IAB specification.

unglud: valid point

src/parser/creative_linear_parser.js
```diff
@@ -158,6 +159,30 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
...
+      const mezzanineElement = parserUtils.childByName(mediaFilesElement, 'Mezzanine');
+      if (mezzanineElement) {
+        const mezzanine = new Mezzanine();
+        mezzanine.id = mezzanineElement.getAttribute('id');
+        mezzanine.fileURL = parserUtils.parseNodeText(mezzanineElement);
+        mezzanine.delivery = mezzanineElement.getAttribute('delivery');
+        mezzanine.codec = mezzanineElement.getAttribute('codec');
+        mezzanine.type = mezzanineElement.getAttribute('type');
+        mezzanine.width = parseInt(
+            mezzanineElement.getAttribute('width') || 0, 10
```
rumesh: I'm not a big fan of having default value at 2 places, here you still put 0 as default value but inside the constructor of `class Mezzanine` you also put 0 as default value. I'm asking do we really need to do the `validate` on the class level ? shouldn't we do an early return as soon as we parsed the Mezzanine Element ?
```suggestion
            if (mezzanineElement && mandatoryFields1 !== null etc ...) {
```
Instead of instantiating `new Mezzanine()`to discard the Mezzanin ?
WDYT ?

unglud: we have 4 mandatory fields,
And
if(!field1 || !field2 || !field3 || !field4) return
is ugly
parseInt(null, 10) would give me NaN and i dont want to check 3 different values (null, 0, NaN) so I stick with 0

Logically everything is fine I hope, and here we discuss just style preferences, I would like to avoid it. Any suggestions?

kobawan: I will be taking over Rumesh as he is off. What he meant is that you should first parse the required fields (without instantiating the class), and then call a function to check if those required fields are valid. If they are valid, you can instantiate the class. If they aren't, we should discard the Mezzanine Element

kobawan: The reason why we want this is because we are moving away from classes. Initially they were only used as a way of documentation, they aren't necessary anymore.

src/parser/creative_linear_parser.js
```diff
@@ -158,6 +159,27 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
 
           creative.mediaFiles.push(mediaFile);
         });
+
+      const mezzanineElement = parserUtils.childByName(mediaFilesElement, 'Mezzanine');
+      if (mezzanineElement &&
+        mezzanineElement.getAttribute('delivery') &&
+        mezzanineElement.getAttribute('type') &&
+        mezzanineElement.getAttribute('width') &&
+        mezzanineElement.getAttribute('height')) {
```
kobawan: I think you should move these conditions to a separate function called i.e. `getMezzanineRequiredAttributes` and inside that function return either the object with the required attributes or null. 
It will make this part of the code more explicit as to what is the condition for.

src/parser/creative_linear_parser.js
```diff
@@ -273,3 +271,24 @@ function parseYPosition(yPosition) {
...
+
+/**
+ * Getting required attributes from element
+ * @param  {Object} element - DOM element
+ * @param  {Array} attributes - list of attributes
+ * @return {Array|null} null if a least one element not present
+ */
+function getRequiredAttributes (element, attributes) {
+  const values = {};
+  const error = attributes.some((name) => {
```
kobawan: IMO `some` should not have mutations inside, it should just return a boolean.
You should split the assertion and the object assignment

src/parser/creative_linear_parser.js
```diff
@@ -273,3 +271,24 @@ function parseYPosition(yPosition) {
...
+    if (!element || element.getAttribute(name) == null || element.getAttribute(name) == 0) {
```
kobawan: what about `undefined` or `NaN`?

src/parser/creative_linear_parser.js
```diff
@@ -273,3 +271,24 @@ function parseYPosition(yPosition) {
...
+      return true;
+    }
+
+    values[name] = element.getAttribute(name);
+  });
+
+  if (!error) {
+    return values;
```
kobawan: it is best to be descriptive to what this function returns:
```
return !error ? values : undefined;
```

src/parser/creative_linear_parser.js
```diff
@@ -251,3 +271,24 @@ function parseYPosition(yPosition) {
 
   return parseInt(yPosition || 0);
 }
+
+/**
+ * Getting required attributes from element
+ * @param  {Object} element - DOM element
+ * @param  {Array} attributes - list of attributes
+ * @return {Array|null} null if a least one element not present
```
kobawan: it returns an object or undefined actually

kobawan: It should be `Object|null`

kobawan:
@unglud The PR shows conflicts, I can't merge

unglud:
I think it's not conflicts but changes request, no?
https://www.dropbox.com/s/lukee25vnsev7e4/Screenshot%202019-08-27%2014.12.41.png?dl=0

kobawan:
@unglud No it says `This branch cannot be rebased due to conflicts`

unglud:
This github is so broken, should I create new PR with squashed commits?

kobawan:
Yes it is probably best to create a new PR with squashed commits, just a reminder to fix the comment above about the jsdocs :)

PR #307, opened 2019-09-12
[client] Send warning event on missing required fields 

ZacharieTFR:
### Description
Trigger a "VAST-warning" if any required attributes or subelements are missing in any element.
According to https://iabtechlab.com/wp-content/uploads/2018/11/VAST4.1-final-Nov-8-2018.pdf

 ### Type
- [x] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/spec/vast_verification.spec.js
+++ b/spec/vast_verification.spec.js
@@ -212,7 +212,7 @@ describe('parserVerification', function() {
       const emitMock = jest.fn();
       const element = getNodesFromXml('<MediaFile></MediaFile>');
       parserVerification.verifyRequiredAttributes(element, emitMock);
-      expect(emitMock.mock.calls.length).toBe(1);
+      expect(emitMock).toHaveBeenCalledTimes(1);
     });
 
     it('should not call the emit function when missing non required attributes', () => {
@@ -222,7 +222,7 @@ describe('parserVerification', function() {
         '<MediaFile delivery="progressive" type="video/mp4" width="1280" height="720"></MediaFile>'
       );
       parserVerification.verifyRequiredAttributes(element, emitMock);
-      expect(emitMock.mock.calls.length).toBe(0);
+      expect(emitMock).toHaveBeenCalledTimes(0);
     });
   });
 
@@ -236,7 +236,7 @@ describe('parserVerification', function() {
         emitMock,
         isAdInLine
       );
-      expect(emitMock.mock.calls.length).toBe(1);
+      expect(emitMock).toHaveBeenCalledTimes(1);
     });
 
     it('should not call the emit function if node inside Wrapper has a missing sub element that is required when inside InLine', () => {
@@ -248,7 +248,7 @@ describe('parserVerification', function() {
         emitMock,
         isAdInLine
       );
-      expect(emitMock.mock.calls.length).toBe(0);
+      expect(emitMock).toHaveBeenCalledTimes(0);
     });
   });
 });
--- a/spec/vast_verification.spec.js
+++ b/spec/vast_verification.spec.js
@@ -1,15 +1,11 @@
 import { parserVerification } from '../src/parser/parser_verification';
 import { vastInLine } from './samples/inline_missing_values';
 import { vastMissingWrapperValues } from './samples/wrapper_missing_values';
+import { getNodesFromXml } from './utils/utils';
 
 describe('parserVerification', function() {
   describe('verifyRequiredValues', function() {
     let warnings = [];
-
-    const parseXmlFromString = xmlString => {
-      var parser = new DOMParser();
-      return parser.parseFromString(xmlString, 'application/xml');
-    };
 
     const emitter = (emitType, { message }) => {
       if (emitType === 'VAST-warning') {
@@ -22,7 +18,7 @@ describe('parserVerification', function() {
     });
 
     it('should have emitted warning for InLine "missing required sub elements"', () => {
-      const inLineElement = parseXmlFromString(
+      const inLineElement = getNodesFromXml(
         vastInLine.missingRequiredSubElements
       );
       parserVerification.verifyRequiredValues(inLineElement, emitter);
@@ -34,7 +30,7 @@ describe('parserVerification', function() {
     });
 
     it('should have emitted warnings "missing required sub elements" for InLine childs: Linear, Icons, MediaFiles', () => {
-      const inLineAndChildsElements = parseXmlFromString(
+      const inLineAndChildsElements = getNodesFromXml(
         vastInLine.childMissingRequiredSubElements
       );
       parserVerification.verifyRequiredValues(inLineAndChildsElements, emitter);
@@ -50,7 +46,7 @@ describe('parserVerification', function() {
     });
 
     it('should have emitted warning "missing required sub elements" for Wrapper', () => {
-      const wrapperElement = parseXmlFromString(vastMissingWrapperValues);
+      const wrapperElement = getNodesFromXml(vastMissingWrapperValues);
       parserVerification.verifyRequiredValues(wrapperElement, emitter);
       expect(warnings).toEqual(
         expect.arrayContaining([
@@ -59,7 +55,7 @@ describe('parserVerification', function() {
       );
     });
     it('should have emitted warning "missing required attributes" for Wrapper', () => {
-      const wrapperElement = parseXmlFromString(vastMissingWrapperValues);
+      const wrapperElement = getNodesFromXml(vastMissingWrapperValues);
       parserVerification.verifyRequiredValues(wrapperElement, emitter);
       expect(warnings).toEqual(
         expect.arrayContaining([
@@ -69,7 +65,7 @@ describe('parserVerification', function() {
     });
 
     it('should have emitted warnings "missing attributes" and warnings "missing one of following subElements"', () => {
-      const missingAttributesElement = parseXmlFromString(
+      const missingAttributesElement = getNodesFromXml(
         vastInLine.missingRequiredAttributes
       );
       parserVerification.verifyRequiredValues(
@@ -97,7 +93,7 @@ describe('parserVerification', function() {
     });
 
     it('should have emitted warnings "Element is empty"', () => {
-      const missingAttributesElement = parseXmlFromString(
+      const missingAttributesElement = getNodesFromXml(
         vastInLine.emptyRequiredValues
       );
       parserVerification.verifyRequiredValues(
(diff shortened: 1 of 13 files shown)


src/parser/ad_parser.js
```diff
@@ -27,29 +28,26 @@ export function parseAd(adElement) {
 
     parserUtils.copyNodeAttribute('id', adElement, adTypeElement);
     parserUtils.copyNodeAttribute('sequence', adElement, adTypeElement);
-
-    if (adTypeElement.nodeName === 'Wrapper') {
-      return parseWrapper(adTypeElement);
-    } else if (adTypeElement.nodeName === 'InLine') {
-      return parseInLine(adTypeElement);
-    }
+    return parseAdElement(adTypeElement, emit);
```
kobawan: I'm not sure what we gained by refactoring this part. Previously parsing inlines and wrappers had a more separate logic, and now parseAdElement does everything. To me it was clearer how inlines and wrappers were parsed previously. Let's discuss this :)

src/parser/creative_companion_parser.js
```diff
@@ -41,10 +41,12 @@ export function parseCreativeCompanion(creativeElement, creativeAttributes) {
         .childrenByName(companionResource, 'StaticResource')
         .reduce((urls, resource) => {
           const url = parserUtils.parseNodeText(resource);
+          const creativeType = resource.getAttribute('creativeType') || null;
```
kobawan: there was no need for this change, now `creativeType` constant is created on every loop of reduce

src/parser/creative_linear_parser.js
```diff
@@ -18,10 +18,10 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
   let offset;
   const creative = new CreativeLinear(creativeAttributes);
 
+  const durationElement = parserUtils.childByName(creativeElement, 'Duration');
```
kobawan: I would undo these changes in the file as well, as they are not needed anymore

src/util/requiredValues.js
```diff
@@ -0,0 +1,64 @@
...
+    subElements: ['MediaFile']
+  },
+  MediaFile: {
+    attributes: ['delivery', 'type', 'width', 'height']
+  },
+  Mezzanine: {
+    attributes: ['delivery', 'type', 'width', 'height']
+  },
+  NonLinear: {
+    inLineRessources: ['StaticResource', 'IFrameResource', 'HTMLResource'],
```
kobawan: typo `inLineRessources`

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ * @param  {Boolean} [isAdInline] - True if node is contained inside a InLine node.
+ *
+ */
+function verifyRequiredValues(node, emit, isAdInline) {
+  if (!node || !node.nodeName) {
+    return;
+  }
+  if (node.nodeName === 'InLine') {
+    isAdInline = true;
```
kobawan: why are we passing `isAdInline` as a param to the function if you're checking for it here?

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+  verifyRequiredAttributes(node, emit);
+
+  if (hasSubElements(node)) {
+    verifyRequiredSubElements(node, emit, isAdInline);
+    for (const nodeKey in node.children) {
```
kobawan: I'm not sure but have you checked that `for...in` doesn't loop also through keys that you don't need. AFAIK `for...in` also loops from its constructor's prototype: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+      verifyRequiredValues(node.children[nodeKey], emit, isAdInline);
+    }
+  } else if (parserUtils.parseNodeText(node).length === 0) {
+    emitMissingValueWarning({ node }, emit);
+    return;
+  }
+
+  return;
```
kobawan: no need to return here :)

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+  verifyRequiredAttributes(node, emit);
+
+  if (hasSubElements(node)) {
+    verifyRequiredSubElements(node, emit, isAdInline);
+    for (const nodeKey in node.children) {
+      verifyRequiredValues(node.children[nodeKey], emit, isAdInline);
+    }
+  } else if (parserUtils.parseNodeText(node).length === 0) {
+    emitMissingValueWarning({ node }, emit);
+    return;
```
kobawan: same as other comment

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+ * @param  {Node} node - The node element
+ * @param  {Boolean} isAdInline - True if node is contained in a inline
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ */
+function verifyRequiredSubElements(node, emit, isAdInline) {
+  const required = requiredValues[node.nodeName];
+  if (!required) {
+    return;
+  }
+  const requiredSubElements = required.subElements;
```
kobawan: this could be undefined as well, as not every requiredValues from the config has subElements mentioned

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Verify and trigger warnings if node required attributes are not set.
+ * @param  {Node} node - The node element.
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ */
+function verifyRequiredAttributes(node, emit) {
+  if (!requiredValues[node.nodeName]) {
+    return;
+  }
+  const requiredAttributes = requiredValues[node.nodeName].attributes;
```
kobawan: this could be undefined

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+    if (!subElement) {
+      emitMissingValueWarning(
+        { node, name: subElementName, type: 'sub element' },
+        emit
+      );
+    }
+  }
+
+  // When InLine format is used some node require at least one resource (i.e <NonLinear>, <Companion>, or <Icon>)
+  if (!isAdInline || !required.inLineRessources) {
```
kobawan: typo!

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+    return;
+  }
+
+  let ressourceFound = false;
+  for (const ressourceKey in required.inLineRessources) {
+    const ressource = required.inLineRessources[ressourceKey];
+    if (parserUtils.childByName(node, ressource)) {
+      ressourceFound = true;
```
kobawan: I don't understand this block, you should make sure every resource in inLineResources is present, not just the first resource found

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+ * Check if a node is sub elements.
+ * @param  {Node} node - The node element.
+ * @returns {Boolean}
+ */
+function hasSubElements(node) {
+  if (
+    !node.children ||
+    node.children.length === 0 ||
+    (node.children.length === 1 &&
+      ['#text', '#cdata-section'].indexOf(node.children[0].nodeName) >= 0)
```
kobawan: I don't think this line should be checked here, this function should just check if it has children. Because what happens if this element has multiple children but they all have no `#text` or `#cdata-section`?

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+  return true;
+}
+
+/**
+ * Trigger Warning for if a node is empty or has missing attributes/subelements
+ * @param  {Object} missingElement - Object containing the node with empty value, or with missing attribute/subElement name and type
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ * @emits  VastParser#VAST-warning
+ */
+function emitMissingValueWarning(missingElement, emit) {
```
kobawan: you could already destructure `missingElement`, that way we know what this function expects

src/parser/parser_verification.js
```diff
@@ -0,0 +1,150 @@
...
+    if (parserUtils.childByName(node, ressource)) {
+      ressourceFound = true;
+      break;
+    }
+  }
+  if (!ressourceFound) {
+    emitMissingValueWarning(
+      {
+        node,
+        name: `${required.inLineRessources.join(' or ')}`,
```
kobawan: You should clarify which resource is missing. The warning messages should be explicit

src/parser/parser_verification.js
kobawan: also maybe instead of having `{ node, name, type }` , you could pass `{ name, parentName, attributes, subElements }`. And in the message check if attributes is passed, or if subElements is passed, or neither, and make a message accordingly.

src/parser/ad_parser.js
```diff
@@ -79,7 +77,8 @@ function parseInLine(inLineElement) {
                 case 'Linear':
                   parsedCreative = parseCreativeLinear(
                     creativeTypeElement,
-                    creativeAttributes
+                    creativeAttributes,
+                    emit
```
kobawan: this is not needed anymore

src/parser/ad_parser.js
```diff
@@ -326,10 +329,7 @@ export function _parseAdVerifications(verifications) {
     const verification = new AdVerification();
     const childNodes = verificationNode.childNodes;
 
-    parserUtils.assignAttributes(
-        verificationNode.attributes,
-        verification
-    );
+    parserUtils.assignAttributes(verificationNode.attributes, verification);
```
kobawan: is this needed if you're calling it in line 42?

src/parser/ad_parser.js
ZacharieTFR: I will remove it

src/parser/ad_parser.js
ZacharieTFR: It's not part of my code, Prettier did that 😅

src/parser/parser_verification.js
ZacharieTFR: I'll replace it with forEach

src/parser/parser_verification.js
ZacharieTFR: I'll add the check

src/parser/parser_verification.js
ZacharieTFR: Yep! I'll add the check

src/parser/parser_verification.js
ZacharieTFR: As discussed in person, the iab specs require at least one of the following resources: StaticResource, IFrameResource, HTMLResource for the nodes <NonLinear>, <Companion>, or <Icon> when InLine format is used

src/parser/parser_verification.js
ZacharieTFR: Indeed, i will remove it

src/parser/parser_verification.js
ZacharieTFR: `isAdInline` is used to know if a node is contained inside an InLine. Since it's a recursive function, it can go through multiple childs levels. I need to pass it through recursive iteration to know when a resource requires it ( one of StaticResource, IFrameResource, HTMLResource is required for the nodes NonLinear, Companion, or Icon when InLine format is used). As discussed in person I'll explain it in the JSDoc

src/parser/ad_parser.js
```diff
@@ -116,7 +128,8 @@ function parseInLine(inLineElement) {
 
       case 'AdVerifications':
         ad.adVerifications = _parseAdVerifications(
-          parserUtils.childrenByName(node, 'Verification')
+          parserUtils.childrenByName(node, 'Verification'),
+          emit
```
kobawan: stray emit caught! :D

src/parser/ad_parser.js
```diff
@@ -316,6 +330,7 @@ function _parseExtension(extNode) {
 /**
  * Parses the AdVerifications Element.
  * @param  {Array} verifications - The array of verifications to parse.
+ * @param  {Function} emit - Emit function used to trigger Warning event
```
kobawan: no need

src/parser/parser_verification.js
```diff
@@ -0,0 +1,180 @@
...
+function verifyRequiredAttributes(node, emit) {
+  if (
+    !requiredValues[node.nodeName] ||
+    !requiredValues[node.nodeName].attributes
+  ) {
+    return;
+  }
+  const requiredAttributes = requiredValues[node.nodeName].attributes;
+  const missingAttributes = [];
+  requiredAttributes.forEach(attributeName => {
```
kobawan: if you wanna be supah cool you could use `filter` method :) it is much less verbose

src/parser/parser_verification.js
```diff
@@ -0,0 +1,180 @@
...
+
+  if (!required) {
+    return;
+  }
+
+  if (required.subElements) {
+    const requiredSubElements = required.subElements;
+    const missingSubElements = [];
+
+    requiredSubElements.forEach(subElementName => {
```
kobawan: same here with `filter`

src/parser/parser_verification.js
```diff
@@ -0,0 +1,180 @@
...
+    }
+  }
+
+  // When InLine format is used some nodes (i.e <NonLinear>, <Companion>, or <Icon>)
+  // require at least one of the following resources: StaticResource, IFrameResource, HTMLResource
+  if (!isAdInline || !required.inLineResources) {
+    return;
+  }
+
+  const resourceFound = required.inLineResources.some(resource => {
```
kobawan: are all the InLine subelements here optional, like `StaticResource, IFrameResource, HTMLResource`?

src/parser/parser_verification.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+}
+
+/**
+ * Check if a node has sub elements.
+ * @param  {Node} node - The node element.
+ * @returns {Boolean}
+ */
+function hasSubElements(node) {
+  return !node.children || node.children.length !== 0;
```
kobawan: shouldn't it be the opposite? `return node.children && node.children.length`

src/parser/parser_verification.js
```diff
@@ -0,0 +1,180 @@
...
+    isAdInline = true;
+  }
+  verifyRequiredAttributes(node, emit);
+
+  if (hasSubElements(node)) {
+    verifyRequiredSubElements(node, emit, isAdInline);
+    for (const nodeKey in node.children) {
+      verifyRequiredValues(node.children[nodeKey], emit, isAdInline);
+    }
+  } else if (parserUtils.parseNodeText(node).length === 0) {
```
kobawan: does this take only cdata or text? We don't want comments

src/parser/parser_verification.js
kobawan: as discussed in person, we shoud rename `inLineResources` to `oneOfinLineResources`

src/parser/parser_verification.js
ZacharieTFR: Totally ! I'll will fix it right away!

src/parser/parser_verification.js
kobawan: ah great to know!

src/parser/ad_parser.js
```diff
@@ -27,29 +29,42 @@ export function parseAd(adElement) {
...
+ * Parses an ad type (Inline or Wrapper)
+ * @param  {Object} adTypeElement - The VAST Inline or Wrapper element to parse.
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ * @emits  VASTParser#VAST-warning
  * @return {Ad}
  */
-function parseInLine(inLineElement) {
-  const childNodes = inLineElement.childNodes;
+function parseAdElement(adTypeElement, emit) {
+  parserVerification.verifyRequiredValues(adTypeElement, emit);
```
kobawan: you should do the check for no `emit` here, so you don't have to call this function

src/parser/parser_verification.js
```diff
@@ -0,0 +1,177 @@
...
+ * @param  {Array} missingElement.subElements - The array of missing sub elements
+ * @param  {Array} missingElement.oneOfResources - The array of resources in which at least one must be provided by the element
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ * @emits  VastParser#VAST-warning
+ */
+function emitMissingValueWarning(
+  { name, parentName, attributes, subElements, oneOfResources },
+  emit
+) {
+  if (!emit) {
```
kobawan: as mentioned check should be done way before-hand

src/parser/parser_verification.js
ZacharieTFR: Actually it's not an array it's an HTMLCollection so i can't use a forEach 😞 . I'll  replace for/in with the classical for loop.

spec/vast_verification.spec.js
```diff
@@ -0,0 +1,122 @@
+import { parserVerification } from '../src/parser/parser_verification';
+import { vastInLine } from './samples/inline_missing_values';
+import { vastMissingWrapperValues } from './samples/wrapper_missing_values';
+
+describe('parserVerification', function() {
+  describe('verifyRequiredValues', function() {
+    let warnings = [];
+
+    const parseXmlFromString = xmlString => {
```
kobawan: I would put this under `spec/utils/utils.js`

spec/vast_verification.spec.js
```diff
@@ -0,0 +1,122 @@
+import { parserVerification } from '../src/parser/parser_verification';
+import { vastInLine } from './samples/inline_missing_values';
+import { vastMissingWrapperValues } from './samples/wrapper_missing_values';
+
+describe('parserVerification', function() {
```
kobawan: Right now you tested that you get warnings for all the things you needed, but you didn't add any tests to check that `verifyRequiredValues`, `verifyRequiredAttributes `, `verifyRequiredSubElements `, `hasSubElements `, `emitMissingValueWarning ` individually do what you expect them to do.

spec/vast_verification.spec.js
ZacharieTFR: I removed it, and replace it with the one already in the utils

spec/vast_verification.spec.js
```diff
@@ -0,0 +1,254 @@
...
+      });
+    });
+  });
+
+  describe('verifyRequiredAttributes', function() {
+    it('should call the emit function for missing required attributes', () => {
+      const emitMock = jest.fn();
+      const element = getNodesFromXml('<MediaFile></MediaFile>');
+      parserVerification.verifyRequiredAttributes(element, emitMock);
+      expect(emitMock.mock.calls.length).toBe(1);
```
kobawan: it is best to use native jest fn `toHaveBeenCalledTimes`

src/parser/parser_verification.js
```diff
@@ -69,8 +69,13 @@ function verifyRequiredAttributes(node, emit) {
  */
 function verifyRequiredSubElements(node, emit, isAdInline) {
   const required = requiredValues[node.nodeName];
-
-  if (!required) {
+  // Wrapper child elements have no required subElement. (Only InLine does)
+  if (
+    !required ||
+    (!isAdInline &&
```
kobawan: you should only depend on `isAdInline` here. It is confusing that you have to check for the node name.

ZacharieTFR: As discussed i will use a better naming 😉

src/parser/parser_verification.js
```diff
@@ -0,0 +1,179 @@
...
+ * @param  {Array} missingElement.subElements - The array of missing sub elements
+ * @param  {Array} missingElement.oneOfResources - The array of resources in which at least one must be provided by the element
+ * @param  {Function} emit - Emit function used to trigger Warning event.
+ * @emits  VastParser#VAST-warning
+ */
+function emitMissingValueWarning(
+  { name, parentName, attributes, subElements, oneOfResources },
+  emit
+) {
+  let message = `Element `;
```
clementFrancon: Might be better to put the name here already and use else if,  and else for the last statement, if we can only have one kind of error per message.

Issue #324, opened 2019-09-23
Add id attribute to missing elements


kobawan:
Currently, the id attribute of some elements are being ignored during parsing. They are the following:
- [x] impressionURLTemplates ids
- [x] advertiser id
- [x] companionClickTrackingURLTemplates ids
- [x] iconClickTrackingURLTemplates ids
- [x] nonlinearClickTrackingURLTemplates ids
- [x] videoClickThroughURLTemplate id
- [x] videoClickTrackingURLTemplates ids
- [x] videoCustomClickURLTemplates ids

The structure for arrays containing urls should be changed to, for example:
```
videoClickTrackingURLTemplates = [
  {
    url : 'http://example.com/linear-clicktracking1',
    id : 'id1'
  },
  {
    url : 'http://example.com/linear-clicktracking2',
    id : 'id2'
  }
];
```

For more info see https://github.com/dailymotion/vast-client-js/issues/131
VAST specs: https://iabtechlab.com/wp-content/uploads/2019/06/VAST_4.2_final_june26.pdf
Development should be done based from branch 3.0-version.

rgulrajani:
Working on it.

PR #326, opened 2019-10-02
Fix3.0/issue 324

rgulrajani:
### Description

Changes to add the id attribute to:
 
- impressionURLTemplates ids
-  advertiser id
- companionClickTrackingURLTemplates ids
- iconClickTrackingURLTemplates ids
- nonlinearClickTrackingURLTemplates ids
- videoClickThroughURLTemplate id
- videoClickTrackingURLTemplates ids
- videoCustomClickURLTemplates ids.

### Issue

https://github.com/dailymotion/vast-client-js/issues/324

### Type
- [x] Breaking change
- [ ] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,6 @@
     "sinon": "^2.3.6"
   },
   "dependencies": {
-    "lodash.isequal": "^4.5.0",
     "xmldom": "^0.1.27"
   }
 }
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "sinon": "^2.3.6"
   },
   "dependencies": {
+    "lodash.isequal": "^4.5.0",
     "xmldom": "^0.1.27"
   }
 }
--- a/src/parser/creative_companion_parser.js
+++ b/src/parser/creative_companion_parser.js
@@ -75,11 +75,14 @@ export function parseCreativeCompanion(creativeElement, creativeAttributes) {
           });
       }
 
-      companionAd.companionClickTrackingURLTemplates = parserUtils
+      parserUtils
         .childrenByName(companionResource, 'CompanionClickTracking')
-        .map(clickTrackingElement =>
-          parserUtils.parseNodeText(clickTrackingElement)
-        );
+        .forEach(clickTrackingElement => {
+          companionAd.companionClickTrackingURLTemplates.push({
+            id: clickTrackingElement.getAttribute('id') || null,
+            url: parserUtils.parseNodeText(clickTrackingElement)
+          });
+        });
 
       companionAd.companionClickThroughURLTemplate =
         parserUtils.parseNodeText(
--- a/src/parser/creative_companion_parser.js
+++ b/src/parser/creative_companion_parser.js
@@ -75,13 +75,13 @@ export function parseCreativeCompanion(creativeElement, creativeAttributes) {
           });
       }
 
-      parserUtils
+      companionAd.companionClickTrackingURLTemplates = parserUtils
         .childrenByName(companionResource, 'CompanionClickTracking')
-        .forEach(clickTrackingElement => {
-          companionAd.companionClickTrackingURLTemplates.push({
+        .map(clickTrackingElement => {
+          return {
             id: clickTrackingElement.getAttribute('id') || null,
             url: parserUtils.parseNodeText(clickTrackingElement)
-          });
+          };
         });
 
       companionAd.companionClickThroughURLTemplate =
--- a/src/parser/creative_linear_parser.js
+++ b/src/parser/creative_linear_parser.js
@@ -51,6 +51,8 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
         id: videoClickThroughElement.getAttribute('id') || null,
         url: parserUtils.parseNodeText(videoClickThroughElement)
       };
+    } else {
+      creative.videoClickThroughURLTemplate = null;
     }
 
     parserUtils
--- a/src/parser/creative_linear_parser.js
+++ b/src/parser/creative_linear_parser.js
@@ -42,24 +42,33 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
     'VideoClicks'
   );
   if (videoClicksElement) {
-    creative.videoClickThroughURLTemplate = parserUtils.parseNodeText(
-      parserUtils.childByName(videoClicksElement, 'ClickThrough')
+    const videoClickThroughElement = parserUtils.childByName(
+      videoClicksElement,
+      'ClickThrough'
     );
+    if (videoClickThroughElement) {
+      creative.videoClickThroughURLTemplate = {
+        id: videoClickThroughElement.getAttribute('id') || null,
+        url: parserUtils.parseNodeText(videoClickThroughElement)
+      };
+    }
 
     parserUtils
       .childrenByName(videoClicksElement, 'ClickTracking')
       .forEach(clickTrackingElement => {
-        creative.videoClickTrackingURLTemplates.push(
-          parserUtils.parseNodeText(clickTrackingElement)
-        );
+        creative.videoClickTrackingURLTemplates.push({
+          id: clickTrackingElement.getAttribute('id') || null,
+          url: parserUtils.parseNodeText(clickTrackingElement)
+        });
       });
 
     parserUtils
       .childrenByName(videoClicksElement, 'CustomClick')
       .forEach(customClickElement => {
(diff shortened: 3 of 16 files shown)


src/parser/creative_companion_parser.js
```diff
@@ -75,11 +75,14 @@ export function parseCreativeCompanion(creativeElement, creativeAttributes) {
           });
       }
 
-      companionAd.companionClickTrackingURLTemplates = parserUtils
+      parserUtils
         .childrenByName(companionResource, 'CompanionClickTracking')
-        .map(clickTrackingElement =>
-          parserUtils.parseNodeText(clickTrackingElement)
-        );
+        .forEach(clickTrackingElement => {
```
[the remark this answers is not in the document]
kobawan: also please update function `mergeWrapperAdData` in `src/parser/parser_utils.js`, it uses `companionClickTrackingURLTemplates`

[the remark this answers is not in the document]
kobawan: and `util.joinArrayUnique` doesn't work anymore, as you need to use deep equality to check for duplicate objects.

src/parser/creative_linear_parser.js
```diff
@@ -42,24 +42,33 @@ export function parseCreativeLinear(creativeElement, creativeAttributes) {
...
   );
   if (videoClicksElement) {
-    creative.videoClickThroughURLTemplate = parserUtils.parseNodeText(
-      parserUtils.childByName(videoClicksElement, 'ClickThrough')
+    const videoClickThroughElement = parserUtils.childByName(
+      videoClicksElement,
+      'ClickThrough'
     );
+    if (videoClickThroughElement) {
+      creative.videoClickThroughURLTemplate = {
```
kobawan: should we assign `creative.videoClickThroughURLTemplate = null` if there is no `videoClickThroughElement`? Maybe it would be clearer to the user that it is not a parsing problem, just that there was no value to parse.

src/util/util.js
```diff
@@ -67,6 +67,32 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
...
 
+/**
+ * Extract the url/s from the URLTemplates.
+ *   If the URLTemplates is an array of urls
+ *   If the URLTemplates object has a url property
+ *   If the URLTemplates is a single string
+ *
+ * TODO: Currently the URLTemplates is not an array when only one url is present.
+ *
+ * @param {Array} URLTemplates - An array of tracking url templates.
```
kobawan: the type should be `Array|String`

src/util/util.js
```diff
@@ -67,6 +67,32 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
   return URLs;
 }
 
+/**
+ * Extract the url/s from the URLTemplates.
+ *   If the URLTemplates is an array of urls
+ *   If the URLTemplates object has a url property
+ *   If the URLTemplates is a single string
+ *
+ * TODO: Currently the URLTemplates is not an array when only one url is present.
```
kobawan: I don't think this comment is necessary

src/util/util.js
```diff
@@ -67,6 +67,32 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
...
+function extractURLsFromTemplates(URLTemplates) {
+  const URLs = [];
+  if (Array.isArray(URLTemplates)) {
+    URLTemplates.forEach(URLTemplate => {
+      if (URLTemplate.hasOwnProperty('url')) {
+        URLs.push(URLTemplate.url);
+      } else {
+        URLs.push(URLTemplate);
+      }
+    });
```
kobawan: A suggestion :)
```
return URLTemplates.map(URLTemplate => {
    return URLTemplate.hasOwnProperty('url') ? URLTemplate.url : URLTemplate
}
```

src/parser/creative_linear_parser.js
kobawan: also please be aware these changes break tracking in `vast_tracker.js` : https://github.com/dailymotion/vast-client-js/blob/master/src/vast_tracker.js#L374
The file will have to be updated as well to ensure all the tracking works

src/parser/creative_companion_parser.js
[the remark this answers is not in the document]
rgulrajani: Done

src/parser/creative_linear_parser.js
rgulrajani: Done

src/util/util.js
rgulrajani: Done

src/util/util.js
rgulrajani: Done

src/util/util.js
rgulrajani: Thanks.
Updated

package.json
```diff
@@ -53,6 +53,7 @@
     "sinon": "^2.3.6"
   },
   "dependencies": {
+    "lodash.isequal": "^4.5.0",
```
rumesh: Do we really need to add this dependency ? we try to keep vast-client with fewer decencies as possible. And when I checked the code we don't need to have `lodash.isequal` which do deeper check inside an object. What changed here is just instead of having a url now we have **plain** object with `id` and `url`.
IMO we should write `isTrackingObjectEqual` which is really simple liek this:

```suggestion
    function isTrackingObjectEqual(obj1, obj2) {
    {
        return obj1.id === obj2.id && obj1.url === obj2.url;
    }
```

kobawan: yes I also agree, lodash.isequal is rather heavy. It is best to do our own custom function for our needs

src/util/util.js
```diff
@@ -128,13 +167,37 @@ function joinArrayUnique(arr1 = [], arr2 = []) {
   }, []);
 }
 
+/**
```
kobawan: you can remove `joinArrayUnique` because as far as I know it was only used in one place. But double check :)

src/vast_tracker.js
```diff
@@ -459,7 +459,7 @@ export class VASTTracker extends EventEmitter {
       variables['CONTENTPLAYHEAD'] = this.progressFormatted();
     }
 
-    util.track(URLTemplates, variables, options);
+    util.track(util.extractURLsFromTemplates(URLTemplates), variables, options);
```
kobawan: maybe we could move this inside `util.track` to be safe, in case `util.track` will be called elsewhere

src/util/util.js
```diff
@@ -47,8 +49,9 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
   // RANDOM/random is not defined in VAST 3/4 as a valid macro tho it's used by some adServer (Auditude)
   variables['RANDOM'] = variables['random'] = variables['CACHEBUSTING'];
 
-  for (const URLTemplateKey in URLTemplates) {
-    let resolveURL = URLTemplates[URLTemplateKey];
+  const URLArray = extractURLsFromTemplates(URLTemplates);
```
kobawan: just a minor comment, I would move this to the top of the function, so it is clear we shouldn't use `URLTemplates` directly

package.json
rgulrajani: Removed the dependency and added a custom method for checking equality

src/util/util.js
rgulrajani: Done

src/vast_tracker.js
rgulrajani: Done

src/util/util.js
rgulrajani: Done

src/util/util.js
```diff
@@ -61,10 +62,70 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
...
+}
+
+/**
+ * Returns a boolean after checking if the object exists in the array.
+ *   true - if the object exists, false otherwise
+ *
+ * @param {Object} obj - The object who existence is to be checked.
+ * @param {Array} list - List of objects.
+ */
+function containsObject(obj, list) {
```
kobawan: sorry just another thing to add, the names of these functions shouldn't be so generic because in fact this function is only usable for url templates. same for `joinArrayOfUniqueObjs`.

rgulrajani:
> Almost done! Please also remove the package-lock.json changes :)
Done. Used npm install --save lodash to remove package

src/util/util.js
rgulrajani: Updated to reflect functionality

Issue #317, opened 2019-09-20
Replace some classes for objects for performance improvements


kobawan:
To improve performance, we should replace the following classes by objects:
- [ ] AdExtension
- [ ] AdVerification
- [ ] Ad
- [ ] CompanionAd
- [ ] CreativeCompanion
- [ ] CreativeLinear
- [ ] CreativeNonLinear
- [ ] Creative
- [ ] Icon
- [ ] MediaFile
- [ ] Mezzanine
- [ ] NonLinearAd
- [ ] VASTResponse

If the classes provide default values, we should make sure not to mutate the default values in the objects.
Development should be done based from branch 3.0-version.

PR #327, opened 2019-10-02
replaced classes with objects

shyamks:
### Description

Refactor for performance improvements.

### Issue

Replacing classes with objects. #317 

### Type

- [x] Enhancement


src/ad_extension.js
```diff
@@ -1,16 +1,13 @@
...
-  }
-
-  isEmpty() {
+export function AdExtension() {
+  this.name = null;
+  this.value = null;
+  this.attributes = {};
+  this.children = [];
+  this.isEmpty = function() {
     return (
```
kobawan: this should be moved out to a new function

src/ad_extension.js
```diff
@@ -1,16 +1,13 @@
...
-  constructor() {
-    this.name = null;
-    this.value = null;
-    this.attributes = {};
-    this.children = [];
-  }
-
-  isEmpty() {
+export function AdExtension() {
+  this.name = null;
```
kobawan: we should not assign the values to `this`, because then we need to depend on the context. We should return a new immutable object

src/vast_response.js
```diff
@@ -1,7 +1,5 @@
-export class VASTResponse {
-  constructor() {
-    this.ads = [];
-    this.errorURLTemplates = [];
-    this.version = null;
-  }
+export function VASTResponse() {
```
kobawan: please don't push yarn.lock files

kobawan:
Also please make sure all the documentation .md files are updated

rumesh:
Please note that function myObject() {....} is still 'class' you can just do a new myObject().
Idea of #317 was to replace Class by plain Object which will have functions.

src/vast_response.js
shyamks: Oh yeah. Sorry about that.

src/ad_extension.js
shyamks: Not sure how you would want me to do it. Changing all except AdExtension.

src/companion_ad.js
```diff
@@ -1,22 +1,22 @@
...
-    this.htmlResources = [];
-    this.iframeResources = [];
-    this.adParameters = null;
-    this.xmlEncoded = null;
-    this.altText = null;
-    this.companionClickThroughURLTemplate = null;
-    this.companionClickTrackingURLTemplates = [];
-    this.trackingEvents = {};
-  }
+export function CompanionAd(creativeAttributes = {}) {
```
kobawan: I think we should change the names of all the classes -> functions. They could be something like: `create<Insert name of the class>`, example: `createCompanionAd`

src/ad_extension.js
kobawan: I mean that in this file you could have two functions: one to create the object, and another that checks if the extension object is empty (by receiving the object as a param)

src/companion_ad.js
shyamks: Since there are no functions anymore, there are snippets where `instanceof` is used for some checks. Will need to figure out how to handle these. 🤔

src/ad_extension.js
```diff
@@ -1,16 +1,16 @@
...
+}
 
-  isEmpty() {
-    return (
-      this.value === null &&
-      Object.keys(this.attributes).length === 0 &&
-      this.children.length === 0
-    );
-  }
+export function isEmpty(adExtension) {
```
rumesh: WDYT ?
```suggestion
export function isEmptyExtension(adExtension) {
```

src/parser/ad_parser.js
```diff
@@ -329,20 +329,20 @@ function _parseExtension(extNode) {
   }
 
   // Only return not empty objects to not pollute extentions
-  return ext.isEmpty() ? null : ext;
+  return isEmpty(ext) ? null : ext;
```
rumesh: with the renaming, it will be much more explicit here while reading the code

src/vast_tracker.js
```diff
@@ -65,7 +65,7 @@ export class VASTTracker extends EventEmitter {
     // Nonlinear and companion creatives provide some tracking information at a variation level
     // While linear creatives provided that at a creative level. That's why we need to
     // differentiate how we retrieve some tracking information.
-    if (this.creative instanceof CreativeLinear) {
+    if (this.creative instanceof createCreativeLinear) {
```
rumesh: This will not work, probably need a function next to `createCreativeLinear` which is `isCreativeLinear` and this will check that this plain object contain all linear creative property

src/companion_ad.js
kobawan: I would add an `adType` property, to distinguish the variation. But that it is just a suggestion :)

src/ad_extension.js
kobawan: I also agree it needs renaming

src/parser/ad_parser.js
```diff
@@ -329,20 +329,20 @@ function _parseExtension(extNode) {
...
 }
 
 /**
  * Parses the AdVerifications Element.
  * @param  {Array} verifications - The array of verifications to parse.
- * @return {Array<AdVerification>}
+ * @return {Array<createAdVerification>}
```
kobawan: it doesn't make much sense to assign this as a type, since it is a function. I would do
```
@return {Array<Object>} an array of ad verification objects
```

src/parser/vast_parser.js
```diff
@@ -218,10 +218,10 @@ export class VASTParser extends EventEmitter {
   /**
    * Builds a VASTResponse which can be returned.
    * @param  {Array} ads - An Array of unwrapped ads
-   * @return {VASTResponse}
+   * @return {createVASTResponse}
```
kobawan: same here with the type. also since `createVASTReponse` only has 3 properties, you could pass them in the function. ex:
```
const response = createVASTResponse({
    ads,
    errorURLTemplates: this.getErrorURLTemplates(),
    version: this.vastVersion,
})
```

src/vast_tracker.js
```diff
@@ -25,7 +25,7 @@ export class VASTTracker extends EventEmitter {
    * @param {VASTClient} client - An instance of VASTClient that can be updated by the tracker. [optional]
    * @param {Ad} ad - The ad to track.
    * @param {Creative} creative - The creative to track.
-   * @param {CompanionAd|NonLinearAd} [variation=null] - An optional variation of the creative.
+   * @param {createCompanionAd|createNonLinearAd} [variation=null] - An optional variation of the creative.
```
kobawan: fix the type pls :)

test/vast_parser.js
```diff
@@ -2,10 +2,11 @@ import path from 'path';
 import should from 'should';
 import sinon from 'sinon';
 import { VASTParser } from '../src/parser/vast_parser';
-import { VASTResponse } from '../src/vast_response';
+// import { createVASTResponse } from '../src/vast_response';
```
rumesh: this line should be removed right ?

src/companion_ad.js
```diff
@@ -1,22 +1,27 @@
...
+    adParameters: null,
+    xmlEncoded: null,
+    altText: null,
+    companionClickThroughURLTemplate: null,
+    companionClickTrackingURLTemplates: [],
+    trackingEvents: {}
+  };
+}
+
+export function isCompanionAd(companionAd) {
```
kobawan: you named the param `companionAd` but it might not be one. I would rename it to `ad` or `adVariation`

test/vast_parser.js
```diff
@@ -137,7 +137,7 @@ describe('VASTParser', function() {
     });
 
     it('should have returned a VAST response object', () => {
-      this.response.should.be.an.instanceOf(VASTResponse);
+      isVASTResponse(this.response).should.eql(true);
```
kobawan: I don't think it is a good idea to introduce a function in the production code just for unit tests. You should instead make sure that `this.response` has the correct properties with the correct values explicitly in the unit test.

src/companion_ad.js
kobawan: same for the other functions

PR #328, opened 2019-10-04
Add tool to compare bundle sizes

kobawan:
### Description

It is possible now to compare local bundle size vs last build.

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [x] Tooling

Screenshot 2019-10-04 at 10 11 07

### DIFF
--- a/bundleSize.js
+++ b/bundleSize.js
@@ -18,11 +18,11 @@ Object.entries(oldValues).forEach(([name, size]) => {
   const newSize = newValues[name];
   const delta = newSize - size;
   const sizeColorFg = delta <= 0 ? '\x1b[32m' : '\x1b[31m';
-  const reset = '\x1b[0m';
+  const resetColorFg = '\x1b[0m';
   const nameColorFg = '\x1b[36m';
   // eslint-disable-next-line no-console
   console.log(
-    `${nameColorFg}%s${reset}%s${sizeColorFg}%s${reset}`,
+    `${nameColorFg}%s${resetColorFg}%s${sizeColorFg}%s${resetColorFg}`,
     `${name}:`,
     ` ${size} B -> ${newSize} B `,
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -25,7 +25,9 @@
   "main": "dist/vast-client-node.min.js",
   "browser": "dist/vast-client.min.js",
   "scripts": {
+    "prebuild": "rm -rf dist_old && mkdir dist_old && cp -a dist/. dist_old/",
```
clementFrancon: Not windows (cmd) proof I think but I'm not sure if it's actually a problem

bundleSize.js
```diff
@@ -0,0 +1,30 @@
...
+};
+
+const oldValues = getFileSizes(oldDistPath);
+const newValues = getFileSizes(newDistPath);
+
+Object.entries(oldValues).forEach(([name, size]) => {
+  const newSize = newValues[name];
+  const delta = newSize - size;
+  const sizeColorFg = delta <= 0 ? '\x1b[32m' : '\x1b[31m';
+  const reset = '\x1b[0m';
```
clementFrancon: I spent some minutes try to figure out what this reset variable stood for, It's maybe only because I'm not used to manipulate colors in console outputs but maybe a naming like "resetColorFg" or something like this would helped me a bit.

package.json
kobawan: I'm not sure how I would make it work on both windows and mac 🤔 is it ok if I leave it as is?

bundleSize.js
kobawan: sure I can rename it

package.json
clementFrancon: I guess it's ok like that, there is still ways for windows users to use it, so install the bash from "windows features" (from  what I quickly checked on the internet) or use Git bash (what I do)

Issue #321, opened 2019-09-20
[VAST 4.1] Implement missing sub elements for Creative


kobawan:
According to IAB 4.1 specs, our parser is missing the sub elements for the Creative element:
- [x] UniversalAdId:
       - value: A string identifying the unique creative identifier. Default value is “unknown”
       - required attributes: idRegistry
- [x] CreativeExtensions
       - sub elements: CreativeExtension **

** CreativeExtension:
       - value: Custom XML object
       - attributes: type

For more info on specs, see page 48 of IAB [VAST 4.1 specs](https://iabtechlab.com/wp-content/uploads/2018/11/VAST4.1-final-Nov-8-2018.pdf).
Parsing of creatives are currently being done in https://github.com/dailymotion/vast-client-js/blob/3.0-version/src/parser/ad_parser.js#L64
Development should be done based from branch 3.0-version.

rgulrajani:
Working on it

PR #330, opened 2019-10-08
Fix3.0/issue 321

rgulrajani:
### Description

Implement missing sub elements for Creative as per VAST 4.1 spec

### Issue

https://github.com/dailymotion/vast-client-js/issues/321

### Type
- [ ] Breaking change
- [x] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling


spec/ad_parser.spec.js
```diff
@@ -1,46 +1,57 @@
 import { _parseExtensions } from '../src/parser/ad_parser';
 import { parserUtils } from '../src/parser/parser_utils';
 import { getNodesFromXml } from './utils/utils';
-import { extensions } from './samples/extensions';
+import { adExtensions } from './samples/extensions';
+import { creativeExtensions } from './samples/extensions';
```
kobawan: multiple imports from same location

src/parser/ad_parser.js
```diff
@@ -90,6 +91,16 @@ function parseAdElement(adTypeElement, emit) {
               apiFramework: creativeElement.getAttribute('apiFramework') || null
             };
 
+            const universalAdIdElement = parserUtils.childByName(
```
kobawan: `Creatives` parsing is becoming rather large. If you could, it would be great to move it to a new file in a function.

src/parser/ad_parser.js
```diff
@@ -101,29 +112,42 @@ function parseAdElement(adTypeElement, emit) {
...
-                  }
                   break;
               }
+
+              if (parsedCreative) {
+                if (universalAdIdElement) {
+                  parsedCreative.universalAdId = {
+                    idRegistry: universalAdIdElement.getAttribute('idRegistry'),
+                    value: parserUtils.parseNodeText(universalAdIdElement)
+                  };
```
kobawan: maybe we could set this as a constant already at the top before the for loop, and just assign it to `parsedCreative.universalAdId` when there is `parsedCreative`. This is just so we are not calling these functions on every loop. Same for creative extensions

src/parser/ad_parser.js
```diff
@@ -265,10 +289,10 @@ function parseWrapper(wrapperElement, emit) {
  * @param  {Node[]} extensions - The array of extensions to parse.
  * @return {AdExtension[]} - The nodes parsed to extensions
  */
-export function _parseExtensions(extensions) {
+export function _parseExtensions(extensions, type = 'Ad') {
```
kobawan: please update jsdocs :)

kobawan: also as a bonus, you could also move this function to a new file.

test/vast_parser.js
```diff
@@ -292,6 +292,35 @@ describe('VASTParser', function() {
           linear.duration.should.equal(90.123);
         });
 
+        it('should have a universal ad id', () => {
```
kobawan: we are in the process of migrating unit tests from mocha to jest. Since you are adding new functionality I would prefer it if you make the tests in jest. Same for future PRs ;)

spec/ad_parser.spec.js
```diff
@@ -1,46 +1,57 @@
...
         'Extension'
       );
-      parsedExts = _parseExtensions(extensionNodes);
+      parsedAdExts = _parseExtensions(extensionNodes);
+
+      const creativeExtensionNodes = parserUtils.childrenByName(
+        getNodesFromXml(creativeExtensions),
+        'CreativeExtension'
+      );
+      parsedCreativeExts = _parseExtensions(creativeExtensionNodes, 'Creative');
```
kobawan: I would move this to a new describe block `CreativeExtensions` and rename this as `AdExtensions`. Also as mentioned in the other comment, you could move the new mocha tests here instead.

spec/ad_parser.spec.js
rgulrajani: Fixed

src/parser/ad_parser.js
rgulrajani: Moved to a different file, along with the parsing of Extensions as well.

src/parser/ad_parser.js
rgulrajani: That wont work, since we want to associate the universalAdId with the actual Creative Element. If I move it outside the for, it basically gets associated to another element, since the array creativeElement.childNodes has a bunch of other elements when parsed.

src/parser/ad_parser.js
rgulrajani: Done for both

test/vast_parser.js
rgulrajani: Have added this to the Creatives parser spec and the Extensions parser spec as well.

spec/ad_parser.spec.js
rgulrajani: Moved to a new block.

src/parser/ad_parser.js
kobawan: I think we misunderstood each other, I meant that this could be moved to line 32, like:
```
// right after defining universalAdIdElement
const universalAdId = {
    idRegistry: universalAdIdElement.getAttribute('idRegistry') || 'unknown',
    value: parserUtils.parseNodeText(universalAdIdElement)
}

// ...later inside the for loop

parsedCreative.universalAdId = universalAdId
```

src/ad_extension.js
```diff
@@ -1,16 +1,5 @@
...
+import { createExtension } from './extension';
 
-export function isEmptyExtension(adExtension) {
-  return (
-    adExtension.value === null &&
-    Object.keys(adExtension.attributes).length === 0 &&
-    adExtension.children.length === 0
-  );
+export function createAdExtension() {
+  return createExtension();
```
kobawan: I think that with the latest changes of changing classes to objects have made this file and creative_extension file redundant. We can just call `createExtension` directly.

rgulrajani: done

docs/api/class-reference.md
```diff
@@ -148,13 +153,23 @@ This object represents a generic Creative. It's used as a parent object for more
...
 
 ## AdExtension<a name="ad-extension"></a>
+
 - `name: String|null`
 - `value: any`
 - `attributes: Object`
 - `children: Array<AdExtension>`
 - `isEmpty(): Boolean` Returns true when none of these attributes have a value
 
+## CreativeExtension<a name="creative-extension"></a>
```
kobawan: please update the docs to just have one `Extension` object

rgulrajani: Done

Issue #316, opened 2019-09-20
Implement missing VAST 4.1 trackers


kobawan:
According to IAB specs for [VAST 4.1](https://iabtechlab.com/wp-content/uploads/2018/11/VAST4.1-final-Nov-8-2018.pdf) our vast-client is missing the following trackers:

**Linear and NonLinear Ads Metrics**
- [x] notUsed: This ad was not and will not be played (e.g. it was prefetched for a particular
ad break but was not chosen for playback). This allows ad servers to reuse an ad earlier
than otherwise would be possible due to budget/frequency capping. This is a terminal event; no other tracking events should be sent when this is used. Player support is
optional and if implemented is provided on a best effort basis as it is not technically
possible to fire this event for every unused ad (e.g. when the player itself is terminated
before playback).
- [x] otherAdInteraction: An optional metric that can capture all other user interactions
under one metric such as hover-overs, or custom clicks. It should NOT replace
clickthrough events or other existing events like mute, unmute, pause, etc.

**NonLinear Ads Metrics**
- [x] acceptInvitation: The user clicked or otherwise activated a control used to pause
streaming content, which either expands the ad within the player’s viewable area or
“takes-over” the streaming content area by launching an additional portion of the ad. An ad in video format ad is usually played upon acceptance, but other forms of
media such as games, animation, tutorials, social media, or other engaging media
are also used.
- [x] adExpand: The user activated a control to expand the creative.
- [x] adCollapse: The user activated a control to reduce the creative to its original
dimensions. 
- [x] minimize: The user clicked or otherwise activated a control used to minimize the ad
to a size smaller than a collapsed ad but without fully dispatching the ad from the
player environment. Unlike a collapsed ad that is big enough to display it’s message,
the minimized ad is only big enough to offer a control that enables the user to
redisplay the ad if desired.
- [x] overlayViewDuration: The time that the initial ad is displayed. This time is based on
the time between the impression and either the completed length of display based
on the agreement between transactional parties or a close, minimize, or accept
invitation event.

These events should be implemented in vast-tracker class. All info is in specs mentioned at the top.
Development should be done based from branch 3.0-version.

ZacharieTFR:
I'm on it ! 🏃

Issue #264, opened 2019-01-09
Vast tracker tests refactor


clementFrancon:
The actual vast tracker tests use the vast parser to parse the VAST file "wrapper-a.xml". It would be better to have a mockup of a parsed json to use it instead and then only test the features related to vast tracker without caring about other parts of the project.

PR #333, opened 2019-10-11
Implement missing VAST 4.1 trackers 

ZacharieTFR:
### Description

Implemented following missing trackers:
- notUsed
- otherAdInteraction
- acceptInvitation
- adExpand
- adCollapse
- minimize
- overlayViewDuration

### Issue

[Implement missing VAST 4.1 trackers](https://github.com/dailymotion/vast-client-js/issues/316)

### TODO
- [x] Write unit tests
- [x] Edit documentation
### Type
- [x] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-tracker.md
+++ b/docs/api/vast-tracker.md
@@ -382,3 +382,160 @@ vastTracker.on('creativeView', () => {
   // impression tracking URLs have been called
 });
 ```
+
+### notUsed()
+Must be called if the ad was not and will not be played (e.g. it was prefetched for a particular
+ad break but was not chosen for playback). This allows ad servers to reuse an ad earlier
+than otherwise would be possible due to budget/frequency capping. This is a terminal event;
+no other tracking events should be sent when this is used. Player support is
+optional and if implemented is provided on a best effort basis as it is not technically
+possible to fire this event for every unused ad (e.g. when the player itself is terminated
+before playback). This is a terminal event; no other tracking events should be sent when this is used.
+
+Calls the notUsed tracking URLs.
+
+#### Events emitted
+ * **`notUsed`**
+
+#### Example
+```Javascript
+vastTracker.on('notUsed', () => {
+  // notUsed tracking URLs have been called
+});
+
+// Called notUsed if the ad was not and will not be played.
+vastTracker.notUsed();
+
+```
+
+### otherAdInteraction()
+An optional metric that can capture all other user interactions
+under one metric such as hover-overs, or custom clicks. It should NOT replace
+clickthrough events or other existing events like mute, unmute, pause, etc.
+
+Calls the otherAdInteraction tracking URLs.
+
+#### Events emitted
+ * **`otherAdInteraction`**
+
+#### Example
+```Javascript
+// Bind mouseover listener to the player
+player.addEventListener('mouseover', () => vastTracker.otherAdInteraction() );
+
+vastTracker.on('otherAdInteraction', () => {
+  // otherAdInteraction tracking URLs have been called
+});
+```
+
+### acceptInvitation()
+The user clicked or otherwise activated a control used to pause
+streaming content, which either expands the ad within the player’s viewable area or
+“takes-over” the streaming content area by launching an additional portion of the ad. An ad in video format ad is usually played upon acceptance, but other forms of
+media such as games, animation, tutorials, social media, or other engaging media
+are also used.
+
+Calls the acceptInvitation tracking URLs.
+
+#### Events emitted
+ * **`acceptInvitation`**
+
+#### Example
+```Javascript
+// Bind click listener to the invitation button
+invitationButton.on('click', () => {
+  vastTracker.acceptInvitation();
+});
+
+vastTracker.on('acceptInvitation', () => {
+  // acceptInvitation tracking URLs have been called
+});
+```
+
+### adExpand()
+The user activated a control to expand the creative.
+
+Calls the adExpand tracking URLs.
+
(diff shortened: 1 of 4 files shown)


docs/api/vast-tracker.md
```diff
@@ -375,3 +382,160 @@ vastTracker.on('creativeView', () => {
...
 ```
+
+### notUsed()
+Must be called if the ad was not and will not be played (e.g. it was prefetched for a particular
+ad break but was not chosen for playback). This allows ad servers to reuse an ad earlier
+than otherwise would be possible due to budget/frequency capping. This is a terminal event;
+no other tracking events should be sent when this is used. Player support is
+optional and if implemented is provided on a best effort basis as it is not technically
+possible to fire this event for every unused ad (e.g. when the player itself is terminated
+before playback). This is a terminal event; no other tracking events should be sent when this is used.
```
kobawan: you already mentioned this in line 389 :D

docs/api/vast-tracker.md
```diff
@@ -375,3 +382,160 @@ vastTracker.on('creativeView', () => {
...
+
+Calls the overlayViewDuration tracking URLs.
+
+#### Events emitted
+ * **`overlayViewDuration`**
+
+#### Example
+```Javascript
+// Bind click listener to the ad collapse button
+minimizeButton.on('click', () => {
```
kobawan: copy pasterino!

spec/samples/inline_trackers.js
```diff
@@ -0,0 +1,93 @@
+export const inlineTrackers = `
+<VAST version="2.1">
```
kobawan: why not just add the tracking events to an existing test VAST? like `inlineLinear`?

spec/vast_trackers.spec.js
```diff
@@ -0,0 +1,147 @@
+import { VASTClient } from '../src/vast_client';
+import { VASTParser } from '../src/parser/vast_parser';
+import { VASTTracker } from '../src/vast_tracker';
+import { inlineTrackers } from '../spec/samples/inline_trackers';
+
+const vastParser = new VASTParser();
+const vastClient = new VASTClient();
+
+describe('VASTTracker', function() {
```
kobawan: Ideally for VASTTracker, we should test with already parsed VASTs, see https://github.com/dailymotion/vast-client-js/issues/264.

spec/vast_trackers.spec.js
```diff
@@ -0,0 +1,147 @@
...
+      vastParser.parseVAST(vastXml).then(res => {
+        response = res;
+        done();
+      });
+    });
+
+    describe('#linear', () => {
+      let spyEmitter;
+      let spyTrackUrl;
+      beforeAll(() => {
```
kobawan: IMO this should be a `beforeEach` to avoid unit tests impacting other unit tests

src/vast_tracker.js
```diff
@@ -337,6 +337,85 @@ export class VASTTracker extends EventEmitter {
...
+   * The time that the initial ad is displayed. This time is based on
+   * the time between the impression and either the completed length of display based
+   * on the agreement between transactional parties or a close, minimize, or accept
+   * invitation event.
+   * The time has to be passed using [ADPLAYHEAD] and [MEDIAPLAYHEAD] for VAST 4.1
+   * Calls the overlayViewDuration tracking URLs.
+   * @emits VASTTracker#overlayViewDuration
+   */
+  overlayViewDuration() {
+    this.track('overlayViewDuration');
```
kobawan: did we find out how to send this ad display duration?

ZacharieTFR: Yes through macros but the one we need is deprecated. I will implement #336 and pass duration as a param which will use the new macros

spec/vast_trackers.spec.js
ZacharieTFR: Done !

docs/api/vast-tracker.md
ZacharieTFR: Removed it

src/vast_tracker.js
kobawan: please add a todo comment here then, otherwise it looks finished :)

spec/vast_trackers.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    });
+
+    describe('#minimize', () => {
+      beforeEach(() => {
+        vastTracker.minimize();
+      });
+
+      it('should have emitted minimize event and called trackUrl', () => {
+        expect(spyEmitter).toHaveBeenCalledWith('minimize', {
+          trackingURLTemplates: ['http://example.com/linear-minimize']
```
clementFrancon: Shouldn't this use the value from the ad object instead of hard coding it ? Same for all other hard coded values

ZacharieTFR: True, i replaced it.

kobawan: Actually I disagree with this change, because if one day we change the sample file and accidentally remove `trackingURLTemplates` in the unit test it would pass because `undefined === undefined`. So we could either keep the hardcoded values, or we add another `expect` to make sure the value is not undefined or null. Let's discuss this :)

ZacharieTFR: IMO the second option is better (add another expect). For instance if a macro needs to be added to any URL from the sample file, each hardcoded value will have to be updated. It is not a big deal but if we can avoid that 😅

PR #340, opened 2019-10-18
[tool] Add performance test for parsing time

kobawan:
### Description

Script to calculate the parsing speed of VAST client (fetch and parse) and VAST parser (just parse). Each test is run 100 times, to get the average.
Current alpha 8 results:
Screenshot 2019-11-06 at 15 33 26

### Type
- [x] Breaking change
- [ ] Enhancement
- [ ] Fix
- [ ] Documentation
- [x] Tooling


performance_test.js
```diff
@@ -0,0 +1,40 @@
...
+
+const inlineXml = getXmlFromFilename('inline-linear.xml');
+const VastParser = new VASTParser();
+const numberOfRuns = 500;
+const arr = new Array(numberOfRuns).fill(0);
+
+const promises = arr.map(() => getParsingTime(VastParser, inlineXml));
+
+Promise.all(promises).then(results => {
+  const average = getAverage(results);
```
kobawan: Problem: The higher the `numberOfRuns` the higher the average becomes. I have changed from `Date.now()` to `performance.now()` and it didn't solve the issue. Any inputs here would be helpful :)

kobawan: Solution: Each loop runs the test in a separate process so they don't share resources with other tests and become slower with time. `Promise.all` is also not reliable because it runs all the promises, and only after it has executed all the promises, will it begin to resolve them (so the timing isn't right). Now we wait for each run to be resolved before going to the next one. Downside is that the tests are a bit slow now.

performance/get_time.js
```diff
@@ -0,0 +1,17 @@
...
+const url = path.join('file://', __dirname, '..', 'spec', 'samples', 'inline-linear.xml');
+
+const VastClient = new VASTClient();
+
+const timeBefore = Date.now();
+
+VastClient.get(url)
+  .then(() => {
+    const timeAfter = Date.now();
+    console.log(timeAfter - timeBefore);
```
clementFrancon: not sure if the const declaration is really usefull here

performance/parse_time.js
```diff
@@ -0,0 +1,20 @@
...
+const stringifiedXml = fs.readFileSync(url).toString();
+const inlineXml = new xmldom.DOMParser().parseFromString(stringifiedXml, 'text/xml');
+
+const VastParser = new VASTParser();
+
+const timeBefore = Date.now();
+VastParser.parseVAST(inlineXml)
+  .then(() => {
+    const timeAfter = Date.now();
+    console.log(timeAfter - timeBefore);
```
clementFrancon: same

performance/performance_test.js
```diff
@@ -0,0 +1,46 @@
...
+ * Calculates the average time taken of the given async test by running it multiple times.
+ * Each run is executed in a child process, to not share resources between test runs.
+ * The given async test must pass the time taken in a console log, for the performance test to work.
+ * @param {String} name - name of the test for the console log
+ * @param {String} testFilePath - absolute path to the test file
+ */
+const runAsyncTimeAverageTest = async (name, testFilePath) => {
+  const times = [];
+
+  for(let i = 0; i < numberOfRuns; i++) {
```
clementFrancon:
```suggestion
  for (let i = 0; i < numberOfRuns; i++) {
```

performance/performance_test.js
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      const time = await new Promise((resolve, reject) => {
+        const proc = spawn("node", [testFilePath]);
+
+        proc.stdout.on("data", (data) => {
+          const dataString = data.toString()
+          if (dataString.match(/^\d+|\d+\.\d+$/g)) {
+            resolve(+dataString)
```
clementFrancon: Not a big fan of the unary plus but I guess it's just a matter of habits

performance/performance_test.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {String} name - name of the test for the console log
+ * @param {String} testFilePath - absolute path to the test file
+ */
+const runAsyncTimeAverageTest = async (name, testFilePath) => {
+  const times = [];
+
+  for(let i = 0; i < numberOfRuns; i++) {
+    try {
+      const time = await new Promise((resolve, reject) => {
+        const proc = spawn("node", [testFilePath]);
```
clementFrancon: Candid question, do we really need to go throught a new child process every time ? Just calling a promise that returns the value was not enough ? I recall you talked about the more you run the test the more it took, I guess it's related to that ?

kobawan: maybe now that we don't do `Promise.all` it isn't needed anymore but I will need to test it out. This is safer because we know that each test is not impacting other runs (by sharing the same resources)

clementFrancon: But since you're doing a await they actually running one by one so it won't share anything right ?

kobawan: I tested it without the child processes, and the results changed in correlation to how many runs was performed. So it seems it is still needed

performance/performance_test.js
kobawan: changed it

performance/performance_test.js
kobawan: done

performance/parse_time.js
kobawan: done

performance/get_time.js
kobawan: done

performance/performance_test.js
clementFrancon: If Rumesh or Jakub have an explaination for that I would be happy to know it

rumesh: TL;DR; Yes spawn new process will be faster than doing it in the same execution thread.

> The child_process.spawn() method spawns the child process asynchronously, without blocking the Node.js event loop
https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options
So first of all using `spawn` will not block the node event loop.

And NodeJS is single thread but using `spawn` let the capabilities to use ale the cores in the CPU

> a new process will be created and managed by your OS. That new process can be executed in parallel to the main process as long as your computer has at least 2 virtual CPU cores

https://stackoverflow.com/questions/33381796/does-node-child-process-spawn-run-on-multi-core-cpu

So yes `spawn` will create clean and new thread to execute this test file.

performance/get_time.js
```diff
@@ -0,0 +1,16 @@
+const path = require('path');
+const { VASTClient } = require('../dist/vast-client-node.min.js');
+
+const url = path.join('file://', __dirname, '..', 'spec', 'samples', 'inline-linear.xml');
+
+const VastClient = new VASTClient();
+
+const timeBefore = Date.now();
```
jakub-g: perhaps you could use `process.hrtime()` although it's more complex
https://nodejs.org/api/process.html#process_process_hrtime_time

performance/performance_test.js
jakub-g: spawning `node` for each test feels heavy. Maybe https://nodejs.org/api/vm.html can be useful and more lightweight, although I never used it personally.

performance/get_time.js
```diff
@@ -0,0 +1,16 @@
+const path = require('path');
+const { VASTClient } = require('../dist/vast-client-node.min.js');
+
+const url = path.join('file://', __dirname, '..', 'spec', 'samples', 'inline-linear.xml');
```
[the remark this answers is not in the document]
jakub-g: FWIW the output from `file-url` is like `file:///C:/git/vast-client-js/spec/samples/inline-linear.xml` which when loaded in any browser on Windows works just fine.

Unclear to me why node wants to load `C:\C:\`
It is either issue in `vast-client` or the lib which it uses under the hood to issue requests, or some peculiary of msys/mingw like [this one](https://stackoverflow.com/questions/28533664/how-to-prevent-msys-to-convert-the-file-path-for-an-external-program#28536585)

performance/get_time.js
kobawan: done

performance/get_time.js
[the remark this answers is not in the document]
kobawan: I changed it to using `pathtofileurl`. I wasn't able to test it on windows though, so I'm not sure it will work. I also removed the keyword `__dirname` as it wasn't necessary after all

performance/performance_test.js
kobawan: I looked into it and tried it, but I had some errors (`new vm.Script('node performance/get_time.js')` was throwing an error). As the waiting time for the performance tests to run is "not great, not terrible", and this PR isn't a priority, I will leave it as it is for now, because at least it works 🤷‍♀

performance/get_time.js
```diff
@@ -1,15 +1,16 @@
...
 const VastClient = new VASTClient();
 
-const timeBefore = Date.now();
-
+const timeBefore = process.hrtime();
 VastClient.get(url)
   .then(() => {
-    console.log(Date.now() - timeBefore);
+    const res = process.hrtime(timeBefore);
+    console.log((res[0] * 1e3 + res[1] * 1e-6));
```
jakub-g: I think it should be `1e6` not `1e-6`? (same in the other file)

```
1e6 === 1000000
1e-6 === 0.000001
```
Maybe more readable to just write 1000 / 1000000 instead :)
Or even better, since recent versions of node you can write integers with `_` separators for readability:
`1_000`
`1_000_000`

jakub-g:
Alright so I found the issue on Windows: it's this https://github.com/nodejs/node/issues/24223#issuecomment-436598570

To fix this, this line:
https://github.com/dailymotion/vast-client-js/blob/master/src/urlhandlers/node_url_handler.js#L11
`fs.readFile(url.pathname, ...`
should become:
`fs.readFile(uri.fileURLToPath(url.href), ...`
but it requires node 10.12+ (currently package.json says 8.x). But since master is now at 3.0.0 alpha it's a good time to bump the min version of node required ;)

performance/get_time.js
kobawan: `res[1]` is in nanoseconds so I'm multiplying it by 1e-6, I could also divide it by 1e6, but it is correct as it is. I can change it to be more explicit

kobawan:
@jakub-g thanks for the input, I didn't think to look at node_url_handler for some reason, I thought the problem was in the performance tests.

performance/get_time.js
jakub-g: Ah well, the node docs [suck](https://nodejs.org/api/process.html#process_process_hrtime_time) because they look like the `arr[1]` is a small value while in reality it's huge.

And now I see that you use `ms` while they use `ns` in output. All good 👍

Issue #336, opened 2019-10-15
[Vast 4.1] Implement missing Macros


ZacharieTFR:
According to IAB specs for [VAST 4.1](https://iabtechlab.com/wp-content/uploads/2018/11/VAST4.1-final-Nov-8-2018.pdf) our vast-client is missing macros.

Furthermore the macro `[CONTENTPLAYHEAD]` is now deprecated in VAST 4.1. We should replace it with `[ADPLAYHEAD]` and `[MEDIAPLAYHEAD]`

Following required macros should also be implemented:

- [x] `[REASON]`: Reason code for not executing verification

Optional macros:

- [x] `[ADPLAYHEAD]`: Playhead for ad video or audio. Replaced with the current time offset “HH:MM:SS.mmm”.
- [x] `[MEDIAPLAYHEAD]` : Playhead for the video or audio content (NOT the ad creative). Replaced with the current time offset “HH:MM:SS.mmm” of the underlying video/audio content that the ad is playing within. Not relevant for out-stream ads.
- [x] `[BREAKPOSITION]`:Indicates the position of the ad break within the underlying video/audio content that the ad is playing within.
- [x] `[BLOCKEDADCATEGORIES]`: List of blocked ad categories.  Values must be taken from the
BlockedCategory element. (Currently BlockedAdCategories element is not supported) 
- [x] `[ADCATEGORIES]`: List of desired ad categories
- [x] `[ADCOUNT]`: For VAST requests: the number of ads expected by the player.For tracking pixels: the number of <InLine> ads played within the current chain or tree of VASTs, including the executing one.
- [x] `[TRANSACTIONID]`: An identifier used to correlate a chain of ad requests from the origination (supply) end.
- [x] `[PLACEMENTTYPE]`: Indicates the type of ad placement.
- [x] `[ADTYPE]`: Indicates whether the ad’s intended use case was video, audio, or hybrid, as defined in the adType attribute of the VAST <Ad> element.
- [x] `[UNIVERSALADID]`: Indicates the creative
- [x] `[IFA]`: A resettable advertising ID from a device-specific advertising ID scheme
- [x] `[IFATYPE]`: String value indicating the type of IFA included in the IFA macro
- [x] `[CLIENTUA]`: An identifier of the player and VAST client used.
- [x] `[SERVERUA]`: User-Agent of the server making the request on behalf of a client.
- [x] `[DEVICEUA]`: User-Agent of the device that is rendering the ad to the end user.
[...]

PR #342, opened 2019-10-24
Add missing VAST 4.1 Macros

ZacharieTFR:
### Description

Add support for VAST 4.1 Macros. Macros can now be replaced by their values when trackers are called. 

### Issue
#336 

### TODO
- [x] Unit tests
- [x] Documentation

### Type
- [x] Breaking change
- [x] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling


src/util/macros.js
```diff
@@ -0,0 +1,47 @@
...
+  'LIMITADTRACKING',
+  'REGULATIONS',
+  'GDPRCONSENT',
+  // <BlockedAdCategories> element is not parsed for now so the vastTracker
+  // can't replace the macro with element value automatically.
+  // The player need to pass it inside "macro" parameter when calling trackers
+  'BLOCKEDADCATEGORIES',
+  'ADCATEGORIES', // <Category> element is also not parsed for now. Same instructions as above
+  'ADTYPE', //adType attribute of <Ad> element is not parsed for now. Same instructions as above
+  'ADSERVINGID' // <AdServingId> element is also not parsed for now. Same instructions as above
```
ZacharieTFR: I will open a PR when these elements are supported.

src/util/util.js
```diff
@@ -54,20 +60,62 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
...
     }
-
-    for (const key in variables) {
-      const value = variables[key];
-      const macro1 = `[${key}]`;
-      const macro2 = `%%${key}%%`;
-      resolveURL = resolveURL.replace(macro1, value);
-      resolveURL = resolveURL.replace(macro2, value);
-    }
+    resolveURL = replaceUrlMacros(resolveURL, variables);
```
kobawan: if we could avoid mutations it would be great, because it can make debugging very hard if there is a bug.

src/util/util.js
```diff
@@ -54,20 +60,62 @@ function resolveURLTemplates(URLTemplates, variables = {}, options = {}) {
...
+ */
+function replaceMacrosValues(url, macros) {
+  for (const key in macros) {
+    const value = macros[key];
+    const macro1 = `[${key}]`;
+    const macro2 = `%%${key}%%`;
+    url = url.replace(macro1, value);
+    url = url.replace(macro2, value);
+  }
+  return url;
```
kobawan: as mentioned previously, instead of mutating `url`, this function could return a new url with the changes.

src/vast_tracker.js
```diff
@@ -233,9 +234,9 @@ export class VASTTracker extends EventEmitter {
    * @emits VASTTracker#mute
    * @emits VASTTracker#unmute
    */
-  setMuted(muted) {
+  setMuted(muted, macros = {}) {
```
kobawan: you missed the jsdocs on this one

src/vast_tracker.js
```diff
@@ -368,84 +376,125 @@ export class VASTTracker extends EventEmitter {
...
+   * Must be called if the player did not or was not able to execute the provided
+   * verification code.The [REASON] macro must be filled with reason code
+   * Calls the verificationNotExecuted trackings URLs.
+   *
+   * @param {Object} [macros={}] - An optional Object containing macros and their values to be used and replaced in the tracking calls.
+   * @emits VASTTracker#verificationNotExecuted
+   */
+  verificationNotExecuted(macros = {}) {
+    if (
+      !(this.ad && this.ad.adVerifications && this.ad.adVerifications.length)
```
kobawan: this is a bit hard to read, we could just write:
```
!this.ad || !this.ad.adVerifications || !this.ad.adVerifications.length
```

kobawan: but I'm not sure we should do this check. IMO this should break if you try to call it when there is no `adVerifications`. Or we should send an error. @rumesh what do you think?

src/vast_tracker.js
```diff
@@ -469,8 +518,13 @@ export class VASTTracker extends EventEmitter {
       this.clickThroughURLTemplate || fallbackClickThroughURL;
 
     if (clickThroughURLTemplate) {
+      const timeProgess = this.progressFormatted();
```
kobawan: typo!

src/vast_tracker.js
```diff
@@ -485,9 +539,11 @@ export class VASTTracker extends EventEmitter {
    * Calls the tracking URLs for the given eventName and emits the event.
    *
    * @param {String} eventName - The name of the event.
+   * @param {Object} [variables={}] - An optional Object of parameters(vast macros) to be used in the tracking calls.
    * @param {Boolean} [once=false] - Boolean to define if the event has to be tracked only once.
+   *
    */
-  track(eventName, once = false) {
+  track(eventName, variables, once = false) {
```
kobawan: where is `once` used? Because I don't see the change in the PR. This is a breaking change btw. to avoid the breaking change you should put it last with a default value `{}`. Also all the new params need to be added to the documentation. We should also add there which macros we currently support.

src/vast_tracker.js
```diff
@@ -536,7 +592,29 @@ export class VASTTracker extends EventEmitter {
       ) {
         variables['ASSETURI'] = this.creative.mediaFiles[0].fileURL;
       }
-      variables['CONTENTPLAYHEAD'] = this.progressFormatted();
+
+      const progress = this.progressFormatted();
+      // Do not set contentPlayHead if progress is NaN or already set
+      if (!variables['CONTENTPLAYHEAD'] && progress !== progress) {
```
kobawan: how is `progress !== progress` they are the same variable

src/vast_tracker.js
```diff
@@ -536,7 +592,29 @@ export class VASTTracker extends EventEmitter {
...
+    }
+    if (
+      this.creative &&
+      this.creative.universalAdId &&
+      this.creative.universalAdId.idRegistry &&
+      this.creative.universalAdId.value
+    ) {
+      variables['UNIVERSALADID'] = `${this.creative.universalAdId.idRegistry} ${
+        this.creative.universalAdId.value
+      }`;
```
kobawan: is this according to the specs? 🤔

src/vast_tracker.js
kobawan: Also we should mention which macros we fill automatically, and which ones need to be added by the user.

src/vast_tracker.js
```diff
@@ -368,84 +376,125 @@ export class VASTTracker extends EventEmitter {
...
+   *
+   * @param {String} duration - The time that the initial ad is displayed.
+   * @param {Object} [macros={}] - An optional Object containing macros and their values to be used and replaced in the tracking calls.
    * @emits VASTTracker#overlayViewDuration
    */
-  // TODO : Pass the duration as a parameter and replace the corresponding macro with it
-  overlayViewDuration() {
-    this.track('overlayViewDuration');
+  overlayViewDuration(duration, macros = {}) {
+    macros['CONTENTPLAYHEAD'] = duration;
```
kobawan: btw shouldn't all these macros be encoded with `encodeURIComponentRFC3986`?

src/vast_tracker.js
ZacharieTFR: As discussed in person, i replaced the track optionals parameters with a single object. All the documentation was in work since it was WIP, but now it's done 😉

src/vast_tracker.js
ZacharieTFR: Yes, it's on p.97: `Format:  "registryID idvalue" - with a space separating registryID and value.`

src/vast_tracker.js
ZacharieTFR: They are now all encoded

docs/api/vast-tracker.md
```diff
@@ -535,3 +613,60 @@ vastTracker.on('overlayViewDuration', () => {
...
+});
+// Call the overlayViewDuration event
+vastTracker.verificationNotExecuted({REASON: 3});
+```
+
+### track(eventName, { macros, once })
+Calls the tracking URLs for the given eventName and emits the event.
+
+#### Parameters
+* **`eventName: Object`** - The event name
```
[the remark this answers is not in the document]
ZacharieTFR: Copy pasterinoo 🤦‍♂

src/vast_tracker.js
rumesh: Y6es I agree, this function shouldn't be called if no adVerifications, but to be clean and robuste vast-client should send an error saying no adVerification.
Because if we do not have this check it will be an error like it wasn't handle properly by the vast-client.
So agree on this should break but not because accessing `this.ad.adVerifications` while it's not defined but because we throw an error in this case

ZacharieTFR: Done !

src/util/util.js
```diff
@@ -13,60 +15,96 @@ function track(URLTemplates, variables, options) {
...
+ * @param {Object} macros - Object of macros to be replaced in the tracking calls
+ */
+function replaceMacrosValues(url, macros) {
+  let replacedMacrosUrl = url;
+  for (const key in macros) {
+    const value = macros[key];
+    const macro1 = `[${key}]`;
+    const macro2 = `%%${key}%%`;
+    replacedMacrosUrl = replacedMacrosUrl.replace(macro1, value);
+    replacedMacrosUrl = replacedMacrosUrl.replace(macro2, value);
```
rumesh: Let use power of regex for better performance
```
// this will match [${key}] and %%${key}%%
let macroRegex = new RegExp(`(?:\[|%%)([${key}])(?:\]|%%)`, 'g');
replacedMacrosUrl = replacedMacrosUrl.replace(macroRegex, value);
```

rumesh: Not only better performance btw, with the regexp 'g' option it will replace all the occurence while a normal replace will only replace the first.

ZacharieTFR: Sure, i replaced it. It was a legacy code that i should have optimized before using  😅

Issue #322, opened 2019-09-23
[VAST 4.1] Implement missing attributes and sub elements of Wrapper


kobawan:
According to IAB 4.1 specs, our parser is missing the following for the Wrapper element:
Attributes:
- [ ] followAdditionalWrappers - a Boolean value that identifies whether subsequent Wrappers after a
requested VAST response is allowed. If false, any Wrappers received (i.e. not
an Inline VAST response) should be ignored. Otherwise, VAST Wrappers
received should be accepted (default value is “true.”)
- [ ] allowMultipleAds - a Boolean value that identifies whether multiple ads are allowed in the
requested VAST response. If true, both Pods and stand-alone ads are allowed.
If false, only the first stand-alone Ad (with no sequence values) in the
requested VAST response is allowed. Default value is “false.”
- [ ] fallbackOnNoAd - a Boolean value that provides instruction for using an available Ad when the
requested VAST response returns no ads. If true, the media player should
select from any stand-alone ads available. If false and the Wrapper represents
an Ad in a Pod, the media player should move on to the next Ad in a Pod;
otherwise, the media player can follow through at its own discretion where
no-ad responses are concerned.

Sub Elements:
- [ ] BlockedAdCategories:
- value: A string that provides comma separated list of category codes or labels per
authority that identify the ad content. 
- required attributes: authority

New attributes should be parsed and implemented according to their description.
For more info on specs, see page 82 of IAB VAST 4.1 specs.
Parsing of wrapper is currently being done in https://github.com/dailymotion/vast-client-js/blob/3.0-version/src/parser/ad_parser.js#L164
Development should be done based from branch 3.0-version.

florianorpeliere:
Hey !
I will try to working on it 😸

rumesh:
@florianorpeliere will you have any time to work on this ?

PR #343, opened 2019-10-28
V3.0/issue 322

florianorpeliere:
### Description

Hey!

I tried to implement the behaviours of new Wrapper attributes, be indulgent 😉. I "really" discovering VAST and the project in the same time 😄.

#### `followAdditionalWrappers`
I change the code [here](https://github.com/florianorpeliere/vast-client-js/blob/v3.0/issue-322/src/parser/ad_parser.js#L34). 
In this configuration:
```
Wrapper Ad 1 (followAdditionalWrappers=”false”)
Wrapper Ad 2 (followAdditionalWrappers=”false”)
Inline Ad
``` 
I parse the Wrapper Ad 1 -> I don't parse the Wrapper Ad 2 -> I don't parse Inline Ad.

In this configuration:
```
Wrapper Ad 1 (followAdditionalWrappers=”false”)
Inline Ad
``` 
I parse the Wrapper Ad 1 -> I parse Inline Ad.

If it's ok for you, I will add tests to ensure all is ok 😉 

#### `allowMultipleAds`
I change the code [here](https://github.com/florianorpeliere/vast-client-js/blob/v3.0/issue-322/src/parser/ad_parser.js#L58-L60). 
I put my code in comment because it's a big change in your actual behavior (67 tests failing)
In this configuration:
```
Wrapper Ad (allowMultipleAds=”false”)
Inline Ad
– Ad sequence = 1
– Ad sequence = 2
– Ad standalone
``` 
I parse the Wrapper Ad 1 -> I parse Inline Ad but I keep only the stand alone Ad (no sequence one)

If it's ok for you, I will add tests to ensure all is ok 😉

#### `fallbackOnNoAd`

In my understanding, it's something to implement directly in the player. It's parsed and available on the ad object.

### Documentation

[VAST4.1](https://iabtechlab.com/wp-content/uploads/2018/11/VAST4.1-final-Nov-8-2018.pdf) page 83
[Implementation example](https://varrando.com/may-force-vast-wrapper-options/) with [code example](https://github.com/Varrando/varrando.github.io/tree/master/jw-varrando-player-demo/tests/wrapper-params)

### Issue

Fix https://github.com/dailymotion/vast-client-js/issues/322

### Type
- [x] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### Before merge

* [x] Add tests for parsing + default values
* [x] Add tests for `followAdditionalWrappers`
* [x] Add tests for `allowMultipleAds`
* [x] Add tests for `allowMultipleAds` & `followAdditionalWrappers` in the same time


florianorpeliere:
I will rebase to integrate updates in my branch.

docs/api/class-reference.md
```diff
@@ -186,3 +186,8 @@ This object represents a generic Creative. It's used as a parent object for more
 - `type: String|null` The type of executable resource provided
 - `parameters: String|null`  String intended for bootstrapping the verification code and providing metadata about the current impression
 - `trackingEvents: Object`
+
+## BlockedAdCategories<a name="blocked-ad-categories"></a>
+
+- `attributes: Object`
```
kobawan: Our approach to parsing in the vast-client is to put each attribute as a new property of this object. So in this case, it would be: `authority: String` instead of `attributes: Object`

kobawan: Also this should be mentioned in `Ad` documentation, together with all the new fields added :)

src/parser/ad_parser.js
```diff
@@ -39,17 +49,24 @@ export function parseAd(adElement, emit) {
...
+ * @param  {Object} options - An optional Object of parameters to be used in the parsing process.
  * @emits  VASTParser#VAST-warning
  * @return {Object} ad - The ad object.
  */
-function parseInLine(adElement, emit) {
+function parseInLine(adElement, emit, { allowMultipleAds = false }) {
+  const sequence = adElement.getAttribute('sequence') || null;
+  if (sequence && !allowMultipleAds) {
+    //return;
+  }
```
kobawan: from what I understand, according to the specs ` only the first stand-alone Ad (with no sequence values) in the
requested VAST response is allowed.`. Since we don't know if there are multiple stand-alone ads here, we will also have to do the check once we have all the ads parsed.

src/parser/ad_parser.js
```diff
@@ -143,6 +167,28 @@ function parseAdElement(adTypeElement, emit) {
...
   const ad = parseAdElement(wrapperElement, emit);
+
+  const followAdditionalWrappersValue = wrapperElement.getAttribute(
+    'followAdditionalWrappers'
+  );
+  const allowMultipleAdsValue = wrapperElement.getAttribute('allowMultipleAds');
+  const fallbackOnNoAdValue = wrapperElement.getAttribute('fallbackOnNoAd');
+
+  ad.followAdditionalWrappers =
+    followAdditionalWrappersValue !== ''
```
kobawan: `getAttribute` can return `''` or `null`: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute, so we should check for both here and below

src/parser/parser_utils.js
```diff
@@ -228,6 +228,10 @@ function mergeWrapperAdData(unwrappedAd, wrapper) {
   );
   unwrappedAd.extensions = wrapper.extensions.concat(unwrappedAd.extensions);
 
+  unwrappedAd.followAdditionalWrappers = wrapper.followAdditionalWrappers;
+  unwrappedAd.allowMultipleAds = wrapper.allowMultipleAds;
+  unwrappedAd.fallbackOnNoAd = wrapper.fallbackOnNoAd;
```
kobawan: we are essentially overriding the `unwrappedAd` values with the values of the parent wrapper. I don't think the IAB specs specify what to do in these cases, so at the very least we could leave a comment here that the values from the child wrapper will be overridden. @rumesh what do you think?

test/vast_parser.js
```diff
@@ -882,6 +882,12 @@ describe('VASTParser', function() {
       it('should have carried sequence over from wrapper', () => {
         this.response.ads[0].sequence.should.eql('1');
       });
+
```
kobawan: please add tests in the new stack we have (jest) under spec folder. You should rebase from 3.0-version branch beforehand though because vast_parser.js file was changed a lot and this pr is behind those changes.

test/vastfiles/wrapper-with-blocks.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0"?>
```
kobawan: please move this file to `spec/samples` with the new stack :)

src/parser/ad_parser.js
florianorpeliere: Oops, indeed.

kobawan:
There is one minor problem with `allowMultipleAds` and default value `false`. This means that any VAST that isn't version 4 won't be allowed to receive ad pods, because this isn't backwards compatible 🤔 We will need to discuss it within the team before merging.

src/parser/ad_parser.js
rumesh: `//return;`
why the return is commented ? If we don't need it or if it's for later we should also comment the `if` and put a comment on top of it

src/parser/parser_utils.js
rumesh: sounds good for me

florianorpeliere:
Thanks @rumesh 😉 I will update my pr.

src/parser/vast_parser.js
```diff
@@ -351,20 +368,39 @@ export class VASTParser extends EventEmitter {
...
+        followAdditionalWrappers
       });
     } catch (e) {
       return Promise.reject(e);
     }
 
+    // if wrapper allowMultipleAds is set to false only the first stand-alone Ad
+    // (with no sequence values) in the requested VAST response is allowed
+    if (
+      typeof allowMultipleAds === 'boolean' &&
```
clementFrancon, line 397: We can probably replace `typeof allowMultipleAds === 'boolean' && !allowMultipleAds` by` allowMultipleAds === false`

src/parser/vast_parser.js
```diff
@@ -351,20 +368,39 @@ export class VASTParser extends EventEmitter {
...
+      !allowMultipleAds &&
+      ads.length > 1 &&
+      parseFloat(this.vastVersion) >= 3
+    ) {
+      ads.splice(1);
```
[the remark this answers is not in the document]
ZacharieTFR: It's done! I've also done more manual testing on my end to make sure we cover edge cases.

src/util/macros.js
```diff
@@ -37,11 +37,9 @@ export const supportedMacros = [
...
   'GDPRCONSENT',
-  // <BlockedAdCategories> element is not parsed for now so the vastTracker
-  // can't replace the macro with element value automatically.
-  // The player need to pass it inside "macro" parameter when calling trackers
   'BLOCKEDADCATEGORIES',
   'ADCATEGORIES',
   'ADTYPE',
-  'ADSERVINGID'
+  'ADSERVINGID',
+  'OMIDPARTNER'
```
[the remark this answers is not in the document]
ZacharieTFR: It was the only missing macro to support all of them and as I added the ADSERVINGID macro, it was a good opportunity to add it as well considering how easy it is to do so 😄

PR #365, opened 2020-07-17
[tracker] fix percentage tracker bug

PaulineMDaily:
### Description

The progress event is used to call trackers when the ad duration is equal to or greater than the value provided in an additional offset attribute for the Linear element. Offset values can be time in the format HH:MM:SS or HH:MM:SS.mmm or a percentage value in the format n%. 

For some VAST,  the calculation of the percentages was not accurate when the duration of the ad was short due to `const percent = Math.round((progress / this.assetDuration) * 100);`
Exemple: for an ad that last 10 sec, as each second is equivalent to 10% if we let the ad run for 2 seconds following tracker will be triggered (grossly simplified):
- progress-1 
- progress-10%
- progress-2
- progress-20%

As you can see, with the current calculation some intermediate percentages are missing therefore not called.  
This PR fix it.

### Type
- [ ] Breaking change
- [ ] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/spec/vast_trackers.spec.js
+++ b/spec/vast_trackers.spec.js
@@ -223,5 +223,29 @@ describe('VASTTracker', function() {
         expect(spyEmitter).toHaveBeenCalledTimes(1);
       });
     });
+
+    describe('setProgress', () => {
+      let spyTrack;
+      let lastMock;
+      beforeEach(() => {
+        spyTrack = jest.spyOn(vastTracker, 'track');
+        vastTracker.assetDuration = 10;
+        //  vastTracker.lastDuration = 40;
+        vastTracker.setProgress(5);
+      });
+      it('call track with progress-5', () => {
+        expect(spyTrack).toHaveBeenCalledWith('progress-5', expect.anything());
+      });
+
+      it('call track with progress-50%', () => {
+        expect(spyTrack).toHaveBeenCalledWith(
+          'progress-50%',
+          expect.anything()
+        );
+      });
+      it('should have a lastPercentage variable', () => {
+        expect(vastTracker.lastPercentage).toBeDefined();
+      });
+    });
   });
 });
--- a/spec/vast_trackers.spec.js
+++ b/spec/vast_trackers.spec.js
@@ -226,7 +226,6 @@ describe('VASTTracker', function() {
 
     describe('setProgress', () => {
       let spyTrack;
-      let lastMock;
       beforeEach(() => {
         spyTrack = jest.spyOn(vastTracker, 'track');
         vastTracker.assetDuration = 10;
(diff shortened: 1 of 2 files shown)


spec/vast_trackers.spec.js
```diff
@@ -223,5 +223,43 @@ describe('VASTTracker', function() {
...
+      it('call track with progress-50%', () => {
+        expect(spyTrack).toHaveBeenCalledWith(
+          'progress-50%',
+          expect.anything()
+        );
+      });
+      it('should have a lastPercentage variable', () => {
+        expect(vastTracker.lastPercentage).toBeDefined();
+      });
+      it('should make the lastPercentage variable at the value -1', () => {
```
ZacharieTFR, line 251: I don't understand the goal of this unit test, why should lastPercentage be equal to -1? Also if you want to test a single feature, instead of having multiple `it` with the same title you can do multiple `expect` inside one `it`. But If you want to test the use case where we want to track 5% but detect that the last percentage tracked was 1%, meaning that previous values are missing (2%, 3%, 4%) and should have been triggered you could write:
```js
it('should also calls track for previous missing percentages', () => {
  vastTracker.lastPercentage = 1;
  expect(spyTrack.mock.calls).toContainEqual(
    ['progress-2%', expect.anything()],
    ['progress-3%', expect.anything()],
    ['progress-4%', expect.anything()]
  )
});
```

src/vast_tracker.js
```diff
@@ -37,6 +37,8 @@ export class VASTTracker extends EventEmitter {
     this.impressed = false;
     this.skippable = false;
     this.trackingEvents = {};
+    this.lastPercentage = 0; //We need to keep the last percentage of the tracker in order to
+    //calculate to trigger the events when the VAST duration is short
```
ZacharieTFR, line 41: For better readability, the comment should be moved to the top of the corresponding line.
```suggestion
    this.trackingEvents = {};
    // We need to keep the last percentage of the tracker in order to
    // calculate to trigger the events when the VAST duration is short
    this.lastPercentage = 0; 
```

src/vast_tracker.js
```diff
@@ -178,35 +180,39 @@ export class VASTTracker extends EventEmitter {
...
-          ) {
-            events.push(quartile);
-            this._alreadyTriggeredQuartiles[quartile] = true;
+      const percent = Math.round((progress / this.assetDuration) * 100);
+      for (let i = this.lastPercentage; i < percent; i++) {
+        const events = [];
+        if (progress > 0) {
+          events.push('start');
+          events.push(`progress-${i + 1}%`);
+          events.push(`progress-${Math.round(progress)}`);
```
ZacharieTFR, line 189: As the for loop only concern the percentage progress, we should only push in the events array the missing progress-percentage (`events.push('progress-${i + 1}%');`). Here, we might trigger multiple time 'start' or `progress-${Math.round(progress)}` with the same value, also their is no need to calculate the quartiles inside of it. 
 We need to keep thing same as before but just add your for loop that will push in the events array the missing progress-percentage.

src/vast_tracker.js
clementFrancon: Doesn't look resolved to me 😅

spec/vast_trackers.spec.js
```diff
@@ -223,5 +223,36 @@ describe('VASTTracker', function() {
...
+
+    describe('setProgress', () => {
+      let spyTrack;
+      beforeEach(() => {
+        spyTrack = jest.spyOn(vastTracker, 'track');
+        vastTracker.assetDuration = 10;
+        vastTracker.setProgress(5);
+      });
+      it('should be defined', () => {
+        expect(vastTracker.setProgress).toBeDefined();
```
clementFrancon, line 235: I'm not sure this test is actually usefull since you call the function setProgress in the beforeEach, it would have thrown an error if it was not defined

spec/vast_trackers.spec.js
```diff
@@ -223,5 +223,36 @@ describe('VASTTracker', function() {
...
+      });
+      it('call track with progress-5', () => {
+        expect(spyTrack).toHaveBeenCalledWith('progress-5', expect.anything());
```
clementFrancon, line 238: Do we actually care about the parameter being passed to the function ? Otherwise we can use toHaveBeenCalled() probably. Same with other tests below

ZacharieTFR: For the record, as setProgress doesn't have any unit tests, we do care about the parameter. We must check if tracker are called with the expected values for percentages format/time format.

PR #374, opened 2020-10-14
 Make vastTracker.errorWithCode able to replace more than [ERRORCODE] macro

PaulineMDaily:
### Description

Right now, vastTracker.errorWithCode only take the error code and only replace the [ERRORCODE] macro, we need to improve the function to also handle all other macros.

The ida is to create vastTracker.error() and take a macro object as for all other methods, and deprecate the other method.

### Type
- [ ] Breaking change
- [X] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/spec/vast_trackers.spec.js
+++ b/spec/vast_trackers.spec.js
@@ -282,8 +282,8 @@ describe('VASTTracker', function () {
         expect(spyTrack).not.toHaveBeenCalledTimes(2);
       });
     });
+
     describe('#error', () => {
-      let isCustomCode = false;
       it('should be called with the right arguments', () => {
         vastTracker.error();
         expect(spyTrackUrl).toHaveBeenCalledWith(
@@ -294,14 +294,12 @@ describe('VASTTracker', function () {
       });
     });
 
-    describe('#errorWithCode', () => {
+    fdescribe('#errorWithCode', () => {
       it('should be called with the right arguments', () => {
-        vastTracker.errorWithCode();
-        expect(spyTrackUrl).toHaveBeenCalledWith(
-          ['http://example.com/error_[ERRORCODE]'],
-          expect.objectContaining(expectedMacros),
-          { isCustomCode: false }
-        );
+        const spyError = jest.fn();
+        vastTracker.error = spyError;
+        vastTracker.errorWithCode('1234', true);
+        expect(spyError).toHaveBeenCalledWith({ ERRORCODE: '1234' }, true);
       });
     });
   });
--- a/spec/vast_trackers.spec.js
+++ b/spec/vast_trackers.spec.js
@@ -303,7 +303,6 @@ describe('VASTTracker', function () {
       });
     });
     describe('#errorWithCode', () => {
-      let errorCode = 404;
       beforeEach(() => {
         vastTracker.errorWithCode();
       });
@@ -313,7 +312,7 @@ describe('VASTTracker', function () {
       it('should hava called the error method', () => {
         expect(spyAdError).toHaveBeenCalledTimes(1);
       });
-      it('should have emitted adErrpr, called trackUrl and not emitted any other event', () => {
+      it('should have emitted an ad error, called trackUrl and not emitted any other event', () => {
         expect(spyTrackUrl).toHaveBeenCalledWith(
           ['http://example.com/error_[ERRORCODE]'],
           expect.objectContaining(expectedMacros),
--- a/spec/vast_trackers.spec.js
+++ b/spec/vast_trackers.spec.js
@@ -120,7 +120,7 @@ describe('VASTTracker', function () {
         });
         expect(spyTrackUrl).toHaveBeenCalledWith(
           adTrackingUrls.otherAdInteraction,
-          expect.objectContaining(expectedMacros)
+          expectedMacros
         );
       });
     });
@@ -138,7 +138,7 @@ describe('VASTTracker', function () {
         });
         expect(spyTrackUrl).toHaveBeenCalledWith(
           adTrackingUrls.acceptInvitation,
-          expect.objectContaining(expectedMacros)
+          expectedMacros
         );
       });
     });
@@ -156,7 +156,7 @@ describe('VASTTracker', function () {
         });
         expect(spyTrackUrl).toHaveBeenCalledWith(
           adTrackingUrls.adExpand,
-          expect.objectContaining(expectedMacros)
+          expectedMacros
         );
       });
     });
@@ -174,7 +174,7 @@ describe('VASTTracker', function () {
         });
         expect(spyTrackUrl).toHaveBeenCalledWith(
           adTrackingUrls.adCollapse,
-          expect.objectContaining(expectedMacros)
+          expectedMacros
         );
       });
     });
@@ -219,7 +219,7 @@ describe('VASTTracker', function () {
         });
         expect(spyTrackUrl).toHaveBeenCalledWith(
           adTrackingUrls.notUsed,
-          expect.objectContaining(expectedMacros)
+          expectedMacros
         );
         expect(spyEmitter).toHaveBeenCalledTimes(1);
       });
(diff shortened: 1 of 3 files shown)


src/vast_tracker.js
```diff
@@ -333,7 +333,9 @@ export class VASTTracker extends EventEmitter {
       { isCustomCode }
     );
   }
-
```
ZacharieTFR, line 336: Can you add a JSDoc comment to this function ?

src/vast_tracker.js
```diff
@@ -333,7 +333,9 @@ export class VASTTracker extends EventEmitter {
...
+  error(macros = {}) {
+    this.trackURLs(this.ad.errorURLTemplates, { macros });
```
ZacharieTFR, line 337: trackURLs macro expect a variable containing all the macros not a object with a subobject containing the macros.
e.g if we have the macro DOMAIN='foo.bar', in your case trackURLs  will be called with param `{'macros':{'DOMAIN': 'foo.bar'}}` whereas it just expect  `{'DOMAIN':'foo.bar'}`

test/vast_tracker.js
```diff
@@ -285,18 +285,47 @@ describe('VASTTracker', function() {
         });
       });
 
-      describe('#errorWithCode', () => {
+      describe('#error', () => {
```
ZacharieTFR, line 288: As we are getting rid of the mocha test can you write them in the jest file `vast_trackers.spec.js` please ? 😄

ZacharieTFR (changes requested):
Thanks for the contribution ! 👍 I left few comments

spec/vast_trackers.spec.js
```diff
@@ -13,13 +13,15 @@ describe('VASTTracker', function() {
     let spyTrack;
     let adTrackingUrls;
     let ad;
+    let isCustomCode = false;
```
ZacharieTFR, line 16: This variable doesn't need to be global as it's already defined in the unit test that you have added

spec/vast_trackers.spec.js
```diff
@@ -282,5 +285,41 @@ describe('VASTTracker', function() {
         expect(spyTrack).not.toHaveBeenCalledTimes(2);
       });
     });
+    describe('#error', () => {
+      let isCustomCode = false;
+      beforeEach(() => {
+        vastTracker.error();
+      });
+      it('should be defined', () => {
+        expect(ad.errorURLTemplates).toBeDefined();
```
ZacharieTFR, line 294: You don't need to check if `errorURLTemplates` is defined. The check that you did l.297 is enough 😄

spec/vast_trackers.spec.js
```diff
@@ -282,5 +285,41 @@ describe('VASTTracker', function() {
...
+        );
+        expect(spyTrackUrl).toHaveBeenCalledTimes(1);
+      });
+    });
+    describe('#errorWithCode', () => {
+      beforeEach(() => {
+        vastTracker.errorWithCode();
+      });
+      it('should be defined', () => {
+        expect(ad.errorURLTemplates).toBeDefined();
```
ZacharieTFR, line 310: `errorURLTemplates` have nothing to do with `errorWithCode` function, this assert is not necessary.

spec/vast_trackers.spec.js
```diff
@@ -282,5 +285,41 @@ describe('VASTTracker', function() {
...
+      });
+      it('should hava called the error method', () => {
```
ZacharieTFR, line 312: In this unit test we should check that errorWithCode called error with the expected arguments (macros and isCustomCode) also there is a small typo `hava`

spec/vast_trackers.spec.js
```diff
@@ -282,5 +285,41 @@ describe('VASTTracker', function() {
...
+        expect(spyAdError).toHaveBeenCalledTimes(1);
+      });
+      it('should have emitted an ad error, called trackUrl and not emitted any other event', () => {
+        expect(spyTrackUrl).toHaveBeenCalledWith(
```
ZacharieTFR, line 316: TrackUrl is not called inside the `errorWithCode` function so IMO we should not check that because if so it's more a end to end test. If we check that error is called with the right args it should be enough

src/vast_tracker.js
```diff
@@ -319,19 +319,30 @@ export class VASTTracker extends EventEmitter {
     }
   }
 
+  /**
+   * Send a request to the URI provided by the VAST <Error> element.
+   * @param {Object} [macros={}] - An optional Object containing macros and their values to be used and replaced in the tracking calls.
+   *
+   * @param {Boolean} [isCustomCode=false] - Flag to allow custom values on error code.
+   */
+
```
ZacharieTFR, line 328: Can you remove this breakline ? And the one l.41 🙏 ?

src/vast_tracker.js
```diff
@@ -319,19 +319,30 @@ export class VASTTracker extends EventEmitter {
...
   errorWithCode(errorCode, isCustomCode = false) {
-    this.trackURLs(
-      this.ad.errorURLTemplates,
-      { ERRORCODE: errorCode },
-      { isCustomCode }
-    );
+    this.error({ ERRORCODE: errorCode }, isCustomCode);
+    //eslint-disable-next-line
+    console.log('errorWithCode : this method is deprecated');
```
ZacharieTFR, line 345: What do you think about this? Because it maybe be misunderstood at first glance when seen in the console.
```suggestion
    console.log('The method errorWithCode is deprecated, please use vast tracker error method instead');
```

src/vast_tracker.js
```diff
@@ -319,19 +319,30 @@ export class VASTTracker extends EventEmitter {
...
+   */
+
+  error(macros = {}, isCustomCode = false) {
+    this.trackURLs(this.ad.errorURLTemplates, macros, { isCustomCode });
+  }
+
   /**
    * Send a request to the URI provided by the VAST <Error> element.
    * If an [ERRORCODE] macro is included, it will be substitute with errorCode.
+   * This method is deprecated
```
ZacharieTFR, line 336: You can use the JSDoc `deprecated` flag:  https://jsdoc.app/tags-deprecated.html

ZacharieTFR (changes requested):

spec/vast_trackers.spec.js
```diff
@@ -282,5 +285,34 @@ describe('VASTTracker', function() {
         expect(spyTrack).not.toHaveBeenCalledTimes(2);
       });
     });
+    describe('#error', () => {
+      isCustomCode = false;
+      beforeEach(() => {
+        vastTracker.error();
```
ZacharieTFR, line 291: As there is only one `it` inside this `describe` can we move the error method call inside the `it` ? Also as we use the variable isCustomCode only once IMO we can remove it and do `{isCustomCode: false}` l.297

spec/vast_trackers.spec.js
```diff
@@ -13,13 +13,15 @@ describe('VASTTracker', function() {
     let spyTrack;
     let adTrackingUrls;
     let ad;
+    let spyAdError;
+    let isCustomCode;
```
ZacharieTFR, line 17: As there is only few tests that use the variable `isCustomCode` maybe we could define it there instead of "globally"

spec/vast_trackers.spec.js
clementFrancon: I agree

spec/vast_trackers.spec.js
```diff
@@ -20,20 +20,24 @@ describe('VASTTracker', function () {
...
+      ADPLAYHEAD: '01%3A15%3A05.250',
     };
+
     beforeEach(() => {
       ad = inlineTrackersParsed.ads[0];
       adTrackingUrls = ad.creatives[0].trackingEvents;
       vastTracker = new VASTTracker(vastClient, ad, ad.creatives[0]);
       spyEmitter = jest.spyOn(vastTracker, 'emit');
       spyTrackUrl = jest.spyOn(vastTracker, 'trackURLs');
+      // spyTrackUrl = jest.fn();
```
ZacharieTFR, line 32: You forgot to remove your comments 😄

src/vast_tracker.js
```diff
@@ -650,7 +650,6 @@ export class VASTTracker extends EventEmitter {
         givenMacros['BLOCKEDADCATEGORIES'] = this.ad.blockedAdCategories;
       }
     }
-
     util.track(URLTemplates, givenMacros, options);
```
ZacharieTFR, line 653: Can we keep the line break ? It's better for readability

ZacharieTFR (approved):
LGTM ! 👍

clementFrancon (approved):
LGTM !

Issue #371, opened 2020-10-01
Config for allowMultipleAds


philn5d-ar:
VAST is complex and many providers (demand sources) have mixed interpretation of "sequence" when there is a standalone ad. Even the samples on IAB have sequence in standalone ads. see: https://github.com/InteractiveAdvertisingBureau/VAST_Samples/tree/master/VAST%204.0%20Samples

Google Ad Manager does not seem to have a way to set "allowMultipleAds" for their wrappers either.

It would help to be able to set this to true at the parser lever to mitigate inconsistencies in providers, else wrapped standalone ads will not play when they have a "sequence=1".

ZacharieTFR:
Hello @philn5d-ar,
Indeed as IAB specifiy:

> allowMultipleAds: a Boolean value that identifies whether multiple ads are allowed in the
> requested VAST response. If true, both Pods and stand-alone ads are allowed.
> If false, **only the first stand-alone Ad (with no sequence values) in the
> requested VAST response is allowed**. Default value is “false.”

I recognise that it is an issue when:
- you don't have controls on the value of allowMultipleAds 
- having the default value set to false 
- not accepting ads alone with a sequence attribute knowing that a lot of provider does it and even IAB..

It's a good idea to provide a parameter that we could set a top level inside vastClient/vastParser, please feel free to contribute following our [guidelines](https://github.com/dailymotion/vast-client-js/blob/master/docs/CONTRIBUTING.md#pull-requests) if you want to speed up the process (it's also a great opportunity to participate in [Hacktoberfest 2020](https://hacktoberfest.digitalocean.com/)! 🎉)

ZacharieTFR:
Hello, 
The PR has been merged and is available since the [3.1.0 release](https://github.com/dailymotion/vast-client-js/releases/tag/3.1.0), please update to the latest version and you will be able to set allowMultipleAds value.

As explained in the API documentation, you can set the value using the parameter `option` for following functions:
- VastClient
  - `get(url, options)` [API link](https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-client.md#geturl-options-promise)
- VastParser
  - `getAndParseVAST(url, options) ` [API link](https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-parser.md#getandparsevasturl-options--)
[...]

PR #375, opened 2020-11-04
Fix allowMultipleAds option when resolving wrapper

ZacharieTFR:
### Description

The issue was raised through https://github.com/dailymotion/vast-client-js/issues/371, we can already override the vast value of allowMultipleAds by setting its value inside the option parameter of following public methods:
- VastClient
  - `get(url, options)` [API link](https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-client.md#geturl-options-promise)
- VastParser
  - `getAndParseVAST(url, options) ` [API link](https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-parser.md#getandparsevasturl-options--)
  - `parseVAST(vastXml, options)` [API link](https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-parser.md#getandparsevasturl-options--)

**BUT**
In the context of a wrapper, when resolving, the allowMultipleAds value defined inside the unwrapped vast  was not replaced by the value given in the option. This PR fix it.

### Issue

https://github.com/dailymotion/vast-client-js/issues/371

### Type
- [ ] Breaking change
- [ ] Enhancement
- [X] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-client.md
+++ b/docs/api/vast-client.md
@@ -117,6 +117,8 @@ By default the fully parsed `VASTResponse` contains all the Ads contained in the
   - `urlHandler: URLHandler` - Custom urlhandler to be used instead of the default ones [`urlhandlers`](../../src/urlhandlers)
   - `urlhandler: URLHandler` - Fulfills the same purpose as `urlHandler`, which is the preferred parameter to use
   - `resolveAll: Boolean` - Allows you to parse all the ads contained in the VAST or to parse them ad by ad or adPod by adPod (default `true`)
+  - `allowMultipleAds: Boolean` - A Boolean value that identifies whether multiple ads are allowed in the requested VAST response. This will override any value of allowMultipleAds attribute set in the VAST
+  - `followAdditionalWrappers: Boolean` - a Boolean value that identifies whether subsequent Wrappers after a requested VAST response is allowed. This will override any value of followAdditionalWrappers attribute set in the VAST
 
 #### Example
 
@@ -289,9 +291,11 @@ vastClient.getNextAds(true)
   });
 ```
 
-#### getParser()
+### getParser()
 
 Returns the instance of `VASTParser` used by the client to parse the VAST. Use it to directly call a method provided by the `VASTParser` class.
+
+#### Example
 
 ```Javascript
 const vastClient = new VASTClient();
--- a/src/parser/vast_parser.js
+++ b/src/parser/vast_parser.js
@@ -504,13 +504,8 @@ export class VASTParser extends EventEmitter {
 
       // If allowMultipleAds is set inside the parameter 'option' of public method
       // override the vast value by the one provided
-      let allowMultipleAds;
-      if (this.parsingOptions.allowMultipleAds) {
-        allowMultipleAds = this.parsingOptions.allowMultipleAds;
-      } else {
-        allowMultipleAds = ad.allowMultipleAds;
-      }
-
+      const allowMultipleAds =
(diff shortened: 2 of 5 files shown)


clementFrancon (commented):
A small suggestion otherwise LGTM ;)

src/parser/vast_parser.js
```diff
@@ -497,30 +498,39 @@ export class VASTParser extends EventEmitter {
...
       });
 
+      // If allowMultipleAds is set inside the parameter 'option' of public method
+      // override the vast value by the one provided
+      let allowMultipleAds;
+      if (this.parsingOptions.allowMultipleAds) {
+        allowMultipleAds = this.parsingOptions.allowMultipleAds;
+      } else {
+        allowMultipleAds = ad.allowMultipleAds;
+      }
```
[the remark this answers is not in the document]
ZacharieTFR: Much cleaner ! ✨

docs/api/vast-client.md
```diff
@@ -117,6 +117,8 @@ By default the fully parsed `VASTResponse` contains all the Ads contained in the
   - `urlHandler: URLHandler` - Custom urlhandler to be used instead of the default ones [`urlhandlers`](../../src/urlhandlers)
   - `urlhandler: URLHandler` - Fulfills the same purpose as `urlHandler`, which is the preferred parameter to use
   - `resolveAll: Boolean` - Allows you to parse all the ads contained in the VAST or to parse them ad by ad or adPod by adPod (default `true`)
+  - `allowMultipleAds: Boolean` - A Boolean value that identifies whether multiple ads are allowed in the requested VAST response. This will override any value of allowMultipleAds attribute set in the VAST
```
clementFrancon, line 120: Maybe the doc is misleading, it says it will override any value of allowMultipleAds attribute set in the VAST but actually if I pass allowMultipleAds: false to it and the VAST has allowMultipleAds set to true, we will take the one from the VAST

src/parser/vast_parser.js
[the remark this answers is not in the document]
clementFrancon: Following my last comment we might want to use ?? (Nullish coalescing) operator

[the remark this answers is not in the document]
ZacharieTFR: Done !

docs/api/vast-client.md
ZacharieTFR: Actually, we do want to replace any vast value if allowMultipleAds was provided in the option. Considering the null coalescing operator it will do so 👍

clementFrancon (approved):

clementFrancon:
LGTM 🥇

src/parser/vast_parser.js
```diff
@@ -154,17 +155,17 @@ export class VASTParser extends EventEmitter {
...
-      withCredentials: options.withCredentials
+      withCredentials: options.withCredentials,
     };
-
+    this.maxWrapperDepth = options.wrapperLimit || DEFAULT_MAX_WRAPPER_DEPTH;
+    this.parentURLs = [];
+    this.parsingOptions = { allowMultipleAds: options.allowMultipleAds };
+    this.remainingAds = [];
+    this.rootErrorURLTemplates = [];
+    this.rootURL = '';
```
rumesh, line 168: Any reason why this re order is need? or was it generate by prettier 2.0?

ZacharieTFR: There's no real need, I just thought it would be more readable 🤷‍♂️

Issue #376, opened 2020-11-06
VAST 4.2 allow multiple UniversalAdID elements to be passed in a VAST response


radiantmediaplayer:
Hi,

Great work on this plugin. VAST 4.2 specs states: 

> With VAST 4.2 we now allow multiple UniversalAdID elements to be passed in a VAST response

Right now creative.universalAdId returns an Object `{idRegistry: String, value: String}` in vast-client-js 3.0 and not an Array of Object. If you could add this feature to support multiple UniversalAdID that would be great. You can find a test VAST tag for this at https://github.com/InteractiveAdvertisingBureau/VAST_Samples/blob/master/VAST%204.2%20Samples/Category-test.xml.

Thanks
Arnaud

ZacharieTFR:
Hi Arnaud, 
Thank you for reporting this issue, we will look at it asap ! But if you wish to speed up the process, you can contribute to the repo following our [guidelines](https://github.com/dailymotion/vast-client-js/blob/master/docs/CONTRIBUTING.md#pull-requests).

ZacharieTFR:
Hello, as it's a breaking change the fix has been merged in the [4.X branch](https://github.com/dailymotion/vast-client-js/tree/4.X). We will wait a bit before doing the next major release, in the meantime you can already build and use the [4.X branch](https://github.com/dailymotion/vast-client-js/tree/4.X).
Don't hesitate to contact us if you encounter any difficulties.

PR #377, opened 2020-11-09
Issue-376

rgulrajani:
Updated UniversalAdIds for multiple UniversalAdId nodes

### Description

[VAST-Client] Allow parsing of multiple UniversalAdID

### Issue

https://github.com/dailymotion/vast-client-js/issues/376

### Type
- [x] Breaking change
- [x] Enhancement
- [ ] Fix
- [x] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/class-reference.md
+++ b/docs/api/class-reference.md
@@ -183,7 +183,7 @@ This object represents a generic Creative. It's used as a parent object for more
 
 ## UniversalAdId<a name="universaladid"></a>
 
-- `idRegistry: String|unknown`
+- `idRegistry: String|'unknown'` A string used to identify the URL for the registry website where the unique creative ID is cataloged. Default value is “unknown”.
 - `value: any`
 
 ## AdVerification<a name="ad-verification"></a>
--- a/docs/api/class-reference.md
+++ b/docs/api/class-reference.md
@@ -39,7 +39,7 @@ This object represents a generic Creative. It's used as a parent object for more
 - `adId: String|null`
 - `sequence: Number|null`
 - `apiFramework: String|null`
-- `universalAdId: Object|null`
+- `universalAdIds: Array<Object>` [go to object](#universaladid)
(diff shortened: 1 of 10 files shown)


docs/api/class-reference.md
```diff
@@ -181,6 +181,11 @@ This object represents a generic Creative. It's used as a parent object for more
 - `attributes: Object`
 - `children: Array<Extension>`
 
+## UniversalAdId<a name="universaladid"></a>
+
+- `idRegistry: String|unknown`
+- `value: any`
```
ZacharieTFR, line 187: For `idRegistry` as the default value is a string, I think that we should put `String|"unknown"`  or add the description from IAB specs : `A string used to identify the URL for the registry website where the unique
creative ID is cataloged. Default value is “unknown”.` .
Also `value` is a String

src/vast_tracker.js
```diff
@@ -612,13 +612,12 @@ export class VASTTracker extends EventEmitter {
     }
     if (
       this.creative &&
-      this.creative.universalAdId &&
-      this.creative.universalAdId.idRegistry &&
-      this.creative.universalAdId.value
+      this.creative.universalAdIds &&
+      this.creative.universalAdIds.length
```
ZacharieTFR, line 616: What do you think of using the optional chaining operator for l.617 to 616 ? 
```suggestion
      this.creative?.universalAdIds?.length
```

ZacharieTFR (changes requested):
Thank you for your contribution! 👍 I left some comments.

src/vast_tracker.js
rgulrajani: Updated

clementFrancon (approved):
LGTM ! Great job and thanks for contributing !

ZacharieTFR (approved):

rumesh (approved):
👍🏽

PR #383, opened 2020-12-08
Updated the readme to add new public error method to vastTracker API doc

PaulineMDaily:
### Description

Updated the documentation of vast-tracker to add the new public method: error and add an indication that errorWithCode is deprecated.

### Type
- [ ] Breaking change
- [ ] Enhancement
- [ ] Fix
- [X] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-tracker.md
+++ b/docs/api/vast-tracker.md
@@ -139,7 +139,7 @@ Macros will be replaced and the skip tracking URL will be called with the follow
 ### error(macros, isCustomCode)
 
 Send a request to the URI provided by the VAST <Error> element. `macros` is an optional Object containing macros and their values to be used and replaced in the tracking calls.
-Pass `isCustomCode` as true in order to use any value. If false or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between 100 and 999 (see <https://gist.github.com/rhumlover/5747417>). Otherwise 900 will be used.
+Pass `isCustomCode` as true in order to use any value, `isCustomCode` and its value are related to the[ERRORCODE] macro. If no [ERRORCODE] macro is provided, `isCustomCode`, has no purpose. If false or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between 100 and 999 (see <https://gist.github.com/rhumlover/5747417>). Otherwise 900 will be used.
 
 #### Parameters
 
--- a/docs/api/vast-tracker.md
+++ b/docs/api/vast-tracker.md
@@ -136,6 +136,27 @@ Macros will be replaced and the skip tracking URL will be called with the follow
 
 ## Public Methods 💚 <a name="methods"></a>
 
+### error(macros, isCustomCode)
+
+Send a request to the URI provided by the VAST <Error> element. `macros` is an optional Object containing macros and their values to be used and replaced in the tracking calls.
+Pass `isCustomCode` as true in order to use any value. If false or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between 100 and 999 (see <https://gist.github.com/rhumlover/5747417>). Otherwise 900 will be used.
+
+#### Parameters
+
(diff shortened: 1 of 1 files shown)


docs/api/vast-tracker.md
```diff
@@ -136,7 +136,28 @@ Macros will be replaced and the skip tracking URL will be called with the follow
...
+```Javascript
+const MEDIAFILE_PLAYBACK_ERROR = '405';
+
+// Bind error listener to the player
+player.on('error', () => {
+  vastTracker.error(MEDIAFILE_PLAYBACK_ERROR);
+});
+```
+
+### errorWithCode(errorCode, isCustomCode) : this method is deprecated in favor of the error method
```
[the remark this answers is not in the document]
ZacharieTFR: Actually, we could simplify the link to `[error method](#errormacros-iscustomcode)` IMO it's better to link the current file rather than a frozen version of a commit and more future-proof

docs/api/vast-tracker.md
```diff
@@ -136,10 +136,35 @@ Macros will be replaced and the skip tracking URL will be called with the follow
 
 ## Public Methods 💚 <a name="methods"></a>
 
-### errorWithCode(errorCode, isCustomCode)
+### error(macros, isCustomCode)
 
-Sends a request to the URI provided by the VAST `<Error>` element. If an `[ERRORCODE]` macro is included, it will be substituted with `errorCode`.
-Pass `isCustomCode` as true in order to use any value. If false or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between 100 and 999 (see <https://gist.github.com/rhumlover/5747417>). Otherwise 900 will be used.
+Send a request to the URI provided by the VAST <Error> element. `macros` is an optional Object containing macros and their values to be used and replaced in the tracking calls.
+Pass `isCustomCode` as true in order to use any value, `isCustomCode` and its value are related to the[ERRORCODE] macro. If no [ERRORCODE] macro is provided, `isCustomCode`, has no purpose. If false or no value is passed, the macro will be replaced using `errorCode` only if the code is a number between 100 and 999 (see <https://gist.github.com/rhumlover/5747417>). Otherwise 900 will be used.
```
ZacharieTFR, line 142: errorCode is not a parameter of this method, WDYT about this suggestion ? 
```suggestion
Pass `isCustomCode` as true only if you want to use a custom code for the `[ERRORCODE]` macro. If false, and a value is provided for `[ERRORCODE]` macro the macro will be replaced in the tracking call only if the code is a number between 100 and 999 (see <https://gist.github.com/rhumlover/5747417>). Otherwise 900 will be used.
```

ZacharieTFR (changes requested):

docs/api/vast-tracker.md
```diff
@@ -136,10 +136,35 @@ Macros will be replaced and the skip tracking URL will be called with the follow
...
+  ERRORCODE : customCode 
+}
+
+// Bind error listener to the player
+player.on('error', () => {
+  vastTracker.error(macrosParam);
+});
+```
+
+### errorWithCode(errorCode, isCustomCode) : ⚠️ This method is deprecated in favor of the [error method](https://github.com/dailymotion/vast-client-js/blob/2137a42572da69b1b10b4bbbc30a05dbcceb9083/docs/api/vast-tracker.md#errormacros-iscustomcode)
```
clementFrancon, line 165: Can we add a link (ex: <a name="error"></a>) to the error(macros, isCustomCode) method and use it here as [error method](#error) to be more future proof ?

docs/api/vast-tracker.md
clementFrancon: This change is still missing, right ?

ZacharieTFR (approved):
LGTM 👍

PaulineMDaily:
Thank you for your comments, I have fixed the issues :) Thanks for your approval

Issue #378, opened 2020-11-16
ViewableImpression tags in Wrapper are not collected


radiantmediaplayer:
Hi,

Right now the lib properly handles ViewableImpression tag when in an InLine tag like at: https://www.radiantmediaplayer.com/vast/tags/inline-linear-viewable.xml, but when the ViewableImpression data is in a Wrapper tag it seems the lib is just discarding those information. VAST 4.2 spec indicates that this should be possible to have ViewableImpression in Wrapper and the IAB provides a sample at https://github.com/InteractiveAdvertisingBureau/VAST_Samples/blob/master/VAST%204.0%20Samples/Viewable_Impression-test.xml. 
That would be great if you could take a look into it. 

Arnaud

ZacharieTFR:
Hi Arnaud,
Thank you for reporting this issue, we will look at it asap !

XavM:
Hello @ZacharieTFR,

It seams the Viewable impression tracker is still not firing when in a Wrapper tag.

Any news or eta for a fix ?

Regards,

mouchii:
Hello,

Any update on this ?

ZacharieTFR:
Hello, sorry for the late answer but it has been fixed in the [4.X branch](https://github.com/dailymotion/vast-client-js/tree/4.X). You can use or do a build based on this branch. We will release soon the 4.X branch though

radiantmediaplayer:
Thanks I will look into v4 when it is released.

PR #395, opened 2021-07-16
ViewableImpression wrapper + typo doc

PaulineMDaily:
### Description

As of right now, when the ViewableImpression data is in a Wrapper we are not using the information. VAST 4.2 spec indicates that this should be possible to have ViewableImpression in Wrapper. 
### Issue

https://github.com/dailymotion/vast-client-js/issues/378

### Type
- [x] Breaking change
- [ ] Enhancement
- [X] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/parser/ad_parser.js
+++ b/src/parser/ad_parser.js
@@ -153,7 +153,7 @@ function parseAdElement(adTypeElement, emit) {
         break;
 
       case 'ViewableImpression':
-        ad.viewableImpression = [_parseViewableImpression(node)];
+        ad.viewableImpression.push(_parseViewableImpression(node));
         break;
 
       case 'Description':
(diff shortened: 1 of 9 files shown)


src/parser/ad_parser.js
```diff
@@ -153,7 +153,7 @@ function parseAdElement(adTypeElement, emit) {
         break;
 
       case 'ViewableImpression':
-        ad.viewableImpression = _parseViewableImpression(node);
+        ad.viewableImpression = [_parseViewableImpression(node)];
```
ZacharieTFR, line 156: As we initialized the ad object with an empty array for viewableImpression we can do the same as l.183

ZacharieTFR (changes requested):

clementFrancon (approved):
Same comment as @ZacharieTFR about the push otherwise LGTM

ZacharieTFR (approved):

PR #402, opened 2021-08-24
[parser] Add bitrate estimation from VAST requests

clementFrancon:
### Description

This PR provides a bitrate estimation from the VAST resolution requests.

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/docs/api/vast-parser.md
+++ b/docs/api/vast-parser.md
@@ -334,7 +334,7 @@ vastParser.parseVAST(vastXml, options)
 
 ### getEstimatedBitrate()
 
-Returns the average of the estimated bitrates in bits by secondes.
+Returns the average of the estimated bitrates in kilo bit per secondes.
 
 #### Example
 
@@ -346,7 +346,7 @@ const options = {
 vastParser.parseVAST(vastXml, options)
   .then(res => {
     vastParser.getEstimatedBitrate()
-    // returns 40 000
+    // returns 40
   })
   .catch(err => {
     // Deal with the error
--- a/docs/api/vast-parser.md
+++ b/docs/api/vast-parser.md
@@ -295,8 +295,8 @@ Returns a `Promise` which either resolves with the fully parsed `VASTResponse` o
   - `urlhandler: URLHandler` - Fulfills the same purpose as `urlHandler`, which is the preferred parameter to use
   - `allowMultipleAds: Boolean` - A boolean value that identifies whether multiple ads are allowed in the requested VAST response. This will override any value of allowMultipleAds attribute set in the VAST
   - `followAdditionalWrappers: Boolean` - A boolean value that identifies whether subsequent Wrappers after a requested VAST response is allowed. This will override any value of followAdditionalWrappers attribute set in the VAST
-  - `requestDuration: Number` - The fetching time of the XML. Provide it with byteLength to have a more accurate estimated bitrate.
-  - `byteLength: Number`- The size of the request. Provide it with requestDuration to have a more accurate estimated bitrate.
+  - `requestDuration: Number` - The fetching time of the XML in ms. Provide it with byteLength to have a more accurate estimated bitrate.
+  - `byteLength: Number`- The size of the request in bytes. Provide it with requestDuration to have a more accurate estimated bitrate.
 
 #### Events emitted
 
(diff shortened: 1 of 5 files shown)


docs/api/vast-parser.md
```diff
@@ -330,6 +332,27 @@ vastParser.parseVAST(vastXml, options)
   });
 ```
 
+### getEstimatedBitrate()
+
+Returns the average of the estimated bitrates in bits by secondes.
```
rumesh, line 337:
```suggestion
Returns the average of the estimated bitrates in bytes by secondes.
```

src/parser/vast_parser.js
```diff
@@ -576,4 +580,32 @@ export class VASTParser extends EventEmitter {
...
+   * @param {Number} duration - The duration of the request in ms.
+   */
+  updateEstimatedBitrate(byteLength, duration) {
+    if (!byteLength || !duration || byteLength <= 0 || duration <= 0) {
+      return;
+    }
+
+    // We want the bitrate in bit by seconds, byteLength are in bytes and duration in ms, we need to convert them
+    const bitrate = (byteLength * 8) / duration * 1000;
+    this.estimatedBitrates.push(bitrate);
```
rumesh, line 596:
```suggestion
    const parsedByteLength = parseInt(byteLength, 10);
    const parsedDuration = parseInt(duration, 10);
    if (isNaN(parsedByteLength) || isNaN(parsedDuration) || parsedByteLength <= 0 || parsedDuration <= 0) {
      return;
    }

    // We want the bitrate in bit by seconds, byteLength are in bytes and duration in ms, we need to convert them
    const bitrate = (parsedByteLength * 8) / parsedDuration * 1000;
    this.estimatedBitrates.push(bitrate);
```

WDYT?

Also why not keep as byte rate instead of bit rate?

src/parser/vast_parser.js
```diff
@@ -576,4 +580,32 @@ export class VASTParser extends EventEmitter {
...
+  /**
+   * Returns the estimated bitrate calculated from all previous requests
+   * @returns The average of all estimated bitrates.
+   */
+  getEstimatedBitrate() {
+    if (!this.estimatedBitrates.length) {
+      return 0;
+    }
+
+    const sum = this.estimatedBitrates.reduce((acc, val) => acc + val, 0);
```
rumesh, line 608: Why not do this calculation while adding a new byterate value? and keep the pondaration (total of byterates)?
This way if the estimation is requested several time we could save some calculation? no more array to loop.
WDYT?

docs/api/vast-parser.md
rumesh: ok this comment is releveant only if we switched to byterate 😛

rumesh:
Also when the PR is validated, this need to be on 3.x and 4.x branch also

docs/api/vast-parser.md
```diff
@@ -330,6 +332,27 @@ vastParser.parseVAST(vastXml, options)
...
+
+#### Example
+
+```Javascript
+const options = {
+  requestDuration: 200,
+  byteLength 1000,
```
ZacharieTFR, line 344:
```suggestion
  byteLength: 1000,
```

ZacharieTFR (changes requested):
I left few comments, otherwise LGTM !

spec/bitrate.spec.js
```diff
@@ -0,0 +1,30 @@
+import * as Bitrate from '../src/parser/bitrate';
+
+describe('updateEstimatedBitrate', () => {
+  beforeEach(() => {
+    Bitrate.estimatedBitrate = 0;
+  });
+
+  it('doesn\'t update estimated bitrate if one value is missing', () => {
```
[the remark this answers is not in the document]
clementFrancon: Usually I don't like to switch from simple to double quotes and prefer to stick with the escape because it brings inconsistency. I'll replace with "does not" :trollface:

docs/api/vast-parser.md
clementFrancon: As discussed via PM, we'll stick to the market unit which is kbps

src/parser/vast_parser.js
clementFrancon: done

ZacharieTFR (approved):

PR #403, opened 2021-09-02
[parser] Merge correctly adVerification from extension and from its own node

PaulineMDaily:
### Description

Since VAST 4.1, the adVerification tag isn't in the extensions anymore but has its own node, but it can happen that a VAST contains adVerifications node in the extensions and also in the dedicated AdVerifications node, if that's the case, the vast-parser parse the one from the extension but don't merge it inside the property adVerification of the final parsed ad object, as it should. 

### Type
- [ ] Breaking change
- [ ] Enhancement
- [X] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/spec/ad_parser.spec.js
+++ b/spec/ad_parser.spec.js
@@ -3,7 +3,7 @@ import { getNodesFromXml } from './utils/utils';
 import { parserUtils } from '../src/parser/parser_utils';
 import { linearAd } from './samples/linear_ads';
 import { adVerificationExtensions } from './samples/ad_verification_extentions';
-import { adVerificationFromExtensionsNodeAndFromNode } from './samples/ad_verification_extension_and_node';
+import { adVerificationsInDedicatedNodeAndInExtensionsNode } from './samples/ad_verification_extension_and_node';
 import {
   viewableImpression,
   viewableImpressionPartial,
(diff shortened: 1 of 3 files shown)


src/parser/ad_parser.js
```diff
@@ -116,7 +117,9 @@ function parseAdElement(adTypeElement, emit) {
           from extensions the same way than for an AdVerifications node.
         */
         if (!ad.adVerifications.length) {
-          ad.adVerifications = _parseAdVerificationsFromExensions(extNodes);
+          adVerificationFromExtensions = _parseAdVerificationsFromExensions(
```
ZacharieTFR, line 120: Can you fix this typo ? 
```suggestion
          adVerificationFromExtensions = _parseAdVerificationsFromExtensions(
```

src/parser/ad_parser.js
```diff
@@ -79,6 +79,7 @@ function parseInLine(adElement, emit, { allowMultipleAds } = {}) {
  * @return {Object} ad - The ad object.
  */
 function parseAdElement(adTypeElement, emit) {
+  let adVerificationFromExtensions = [];
```
ZacharieTFR, line 82: It should be in plural 
```suggestion
  let adVerificationsFromExtensions = [];
```

ZacharieTFR (changes requested):
I left minors comments. Can you also update unit tests to cover this fix, pls? 😄

spec/samples/ad_verification_extension_and_node.js
```diff
@@ -0,0 +1,79 @@
+export const adVerificationFromExtensionsNodeAndFromNode = `
```
ZacharieTFR, line 1: This variable name is unclear you should rename it to something like this: `adVerificationsInDedicatedNodeAndInExtensionsNode` , you can also indent the tag for a better readibility and remove unnecessary tags.

spec/ad_parser.spec.js
```diff
@@ -187,12 +188,49 @@ describe('AdParser', function() {
...
+    let adVerificationsFromAd, ad;
+
+    beforeAll(() => {
+      adVerificationsFromAd = getNodesFromXml(
+        adVerificationFromExtensionsNodeAndFromNode
+      );
+      ad = parseAd(adVerificationsFromAd, null).ad;
+    });
+
+    it('should have 4 adVerifications', () => {
```
ZacharieTFR, line 206: The title is not correct, `adVerications` should contains 4 `adVerification` (singular).

ZacharieTFR (changes requested):

ZacharieTFR (approved):

spec/ad_parser.spec.js
PaulineMDaily: Thank you for your comments, I have fixed the issues :) Thanks for your approval

src/parser/ad_parser.js
```diff
@@ -182,12 +185,17 @@ function parseAdElement(adTypeElement, emit) {
...
         ad.blockedAdCategories.push({
           authority: node.getAttribute('authority') || null,
-          value: parserUtils.parseNodeText(node)
+          value: parserUtils.parseNodeText(node),
         });
         break;
     }
   }
 
+  if (adVerificationsFromExtensions.length && ad.system.version < 4.1) {
```
rumesh, line 194: So we stick to the spec and only take adVerification from extension if VAST version is < 4.1
2 question 
- is `system.version` is a number?
- Should we have this restriction with the version (it's an open question)

PaulineMDaily: Hey Rumesh, thank you for your comment, you were right, system.version is a string and as an absolute, we shouldn't leave the browser to deal with the conversion. 

I'll remove this condition anyway because as we are retrocompatible, I don't think we need to be strict with this condition.

ZacharieTFR (approved):

PR #416, opened 2021-11-08
Update xmldom to @xmldom/xmldom

kulakowka:
Since version 0.7.0 this package is published to npm as [@xmldom/xmldom](https://www.npmjs.com/package/@xmldom/xmldom) and no longer as xmldom.


ZacharieTFR:
Done in https://github.com/dailymotion/vast-client-js/pull/419

PR #417, opened 2021-12-07
[fix] Safari cors error 

guest363:
[client][parser] 
When you use in iframe on Safari vastClient can't get VAST. Request is block by CORS.

### Description

- Migrate from xmlhttprequest to fetch. Fetch select between 2 modes: 'cors' : 'no-cors' depends on options.withCredentials
- To macros added support pattern %5B{key}%5D

### Issue

Issue Link [Optional]

### Type
- [ ] Breaking change
- [ ✔️ ] Enhancement
- [ ✔️ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/urlhandlers/xhr_url_handler.js
+++ b/src/urlhandlers/xhr_url_handler.js
@@ -1,61 +1,71 @@
 import { DEFAULT_TIMEOUT } from './consts';
 
-function xhr() {
-  try {
-    const request = new window.XMLHttpRequest();
-    if ('withCredentials' in request) {
-      // check CORS support
-      return request;
-    }
-    return null;
-  } catch (err) {
-    return null;
-  }
-}
-
 function supported() {
-  return !!xhr();
+  return true;
 }
 
-function handleLoad(request, cb) {
-  if (request.status === 200) {
-    cb(null, request.responseXML, {
-      byteLength: request.response.length,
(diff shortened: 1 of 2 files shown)


ZacharieTFR:
Hello, 
Thanks for the contribution, but I think that using a custom `urlHandler` is more appropriate to your need. 
VASTClient's `get` method accepts a custom `URLHandler` as an optional parameter https://github.com/dailymotion/vast-client-js/blob/master/docs/api/vast-client.md#parameters-1
```javascript
const customUrlHandler = {
  get(url, options, cb) {    
     // Add your fetch code with your CORS handling here 
  }
}
const vastClient = new VASTClient();
vastClient.get('http://example.dailymotion.com/vast.xml', options)
  .then(res => {
    // Do something with the parsed VAST response
  })
  .catch(err => {
    // Deal with the error
  });
```
Also I'm not sure to understand the need for `%5B{key}%5D` macro pattern.

guest363:
Macro pattern `%5B` and `%5D` is for support `[` and `]` in url encode. That's actualy a not common case, but i faced this problem.
Custom URLHandler is a good desigion, thats will fix my broblem, thanks.

ZacharieTFR:
Got it 👍 I will close this PR, but feel free to open an issue if you face any other problem !

PR #420, opened 2022-02-11
Ad data on resolving events

clementFrancon:
### Description

In order to troubleshoot ad performances, we need to have some information on the current ad node being resolved 

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/parser/vast_parser.js
+++ b/src/parser/vast_parser.js
@@ -108,13 +108,14 @@ export class VASTParser extends EventEmitter {
    * Fetches a VAST document for the given url.
    * Returns a Promise which resolves,rejects according to the result of the request.
    * @param  {String} url - The url to request the VAST document.
-   * @param {Number} wrapperDepth - how many times the current url has been wrapped
-   * @param {String} previousUrl - url of the previous VAST
+   * @param {Number} wrapperDepth - How many times the current url has been wrapped.
+   * @param {String} previousUrl - Url of the previous VAST.
+   * @param {Object} ad - Previously parsed ad node (Wrapper) related to this fetching.
    * @emits  VASTParser#VAST-resolving
    * @emits  VASTParser#VAST-resolved
    * @return {Promise}
    */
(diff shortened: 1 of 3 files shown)


src/parser/vast_parser.js
```diff
@@ -108,13 +108,14 @@ export class VASTParser extends EventEmitter {
...
-   * @param {String} previousUrl - url of the previous VAST
+   * @param {Number} wrapperDepth - How many times the current url has been wrapped.
+   * @param {String} previousUrl - Url of the previous VAST.
+   * @param {Object} ad - Previously parsed ad node (Wrapper) related to this fetching.
    * @emits  VASTParser#VAST-resolving
    * @emits  VASTParser#VAST-resolved
    * @return {Promise}
    */
-  fetchVAST(url, wrapperDepth = 0, previousUrl = null) {
+  fetchVAST(url, wrapperDepth = 0, previousUrl = null, ad = null) {
```
ZacharieTFR, line 118: What do you think about renaming `ad` to `wrapperAd` to make it more explicit (especially on the markdown documentation)
```suggestion
  fetchVAST(url, wrapperDepth = 0, previousUrl = null, wrapperAd = null) {
```

src/parser/vast_parser.js
```diff
@@ -129,6 +130,7 @@ export class VASTParser extends EventEmitter {
         wrapperDepth,
         maxWrapperDepth: this.maxWrapperDepth,
         timeout: this.fetchingOptions.timeout,
+        ad,
```
clementFrancon, line 133: not sure if we should call it ad or parentAd or something WDYT?

clementFrancon: Wow 1 sec between our comments, ok I'll change it

ZacharieTFR (approved):
LGTM ! 🚀

PR #437, opened 2022-10-27
Add github action for tests and linting

ZacharieTFR:
### Description

Currently we have to run test and check the linting manually. To avoid forgetting to do so I added a GitHub Action at PR creation to run it automatically and to enforce quality.

### Type
- [ ] Breaking change
- [ ] Enhancement
- [ ] Fix
- [ ] Documentation
- [x] Tooling

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [16.x]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [ "master" ]
+
+jobs:
+  validation:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16.x]
```
GuillaumeBAECHLER, line 16: Do you want to keep the matrix if we only check against node 16.x ?

ZacharieTFR: Indeed can be removed !

GuillaumeBAECHLER (approved):
:shipit:

PR #443, opened 2022-12-13
Fix/add muted argument to constructor

GuillaumeBAECHLER:
### Description

The `setMuted` method assumes that `muted` property is false by default.
Thus, the `unmute` tracking event will not fire if the player was in a muted state.

### Changes

This PR adds a way to provide initial mute state through constructor arguments.

### Type
- [ ] Breaking change
- [ ] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling


ZacharieTFR (approved):
LGTM ! 👍

Lotphy (approved):

PR #444, opened 2023-03-14
Add missing node to the parsing 

Rapha0511:
### Description

- The `mergeWrapperAdData` function has been modified to keep track of the <_icon_> elements contained in a wrapper's creative. If a creative contain <_icon_>, the whole creative will be included in the final response.

- Since the VASTTracker allow users to convert seconds to timecode with the `convertToTimecode` function, the `parseDuration` function has been exposed publicly as a VASTParser method to allow users to convert timecode to seconds if needed

- _xmlEncoded_ is now considered as an attribute of the <_adParameters_> node, instead of a node by itself. The value of <_adParameters_> is now an Object of type `{value: String, xmlEncoded: String}`.

- The _type_ attribute of the <_survey_> node has been added to the parsing. The value of survey is now an Object of  type `{value: String, type: String}`.

- <_iconClickFallbackImages_> node has been added to the parsing as an array of Object of type `{url: String, width: String, height: String}`.

- Tests have been added for changes and the addition of the missing nodes. Some of them have been moved from Mocha to Jest

- For companions, the parameter `adSlotID` has been modified to `adSlotId` to follow the VAST specs

[VAST 4.3 features](https://iabtechlab.com/wp-content/uploads/2022/09/VAST_4.3.pdf)

### Type
- [x] Breaking change
- [x] Enhancement
- [ ] Fix
- [x] Documentation
- [ ] Tooling


spec/parser_utils.spec.js
```diff
@@ -385,6 +385,48 @@ describe('ParserUtils', function () {
...
+              width: 1,
+              xPosition: 0,
+              yPosition: 0,
+              apiFramework: 'omid',
+              offset: 5,
+              duration: 12,
+              type: 'image/jpeg',
+              staticResource: 'link',
+              htmlResource:
+                '<!DOCTYPE html>\r\n                                            <html lang="en">\r\n                                            <head>\r\n                                                <meta charset="UTF-8">\r\n                                                <meta http-equiv="X-UA-Compatible" content="IE=edge">\r\n                                                <meta name="viewport" content="width=device-width, initial-scale=1.0">\r\n                                                <title>Document</title>\r\n                                            </head>\r\n                                            <body>\r\n                                                <h1>titre</h1>\r\n                                            </body>\r\n                                            </html>',
```
ZacharieTFR, line 406: Can you replace the htmlResource with a dummy value to make it easier to read ?

spec/parser_utils.spec.js
```diff
@@ -614,6 +656,48 @@ describe('ParserUtils', function () {
...
```
ZacharieTFR, line 677: Same here

src/parser/creative_linear_parser.js
```diff
@@ -303,6 +306,25 @@ function parseIcon(iconElement) {
...
+
+    if (iconClickFallbackImagesElement) {
+      parserUtils
+        .childrenByName(
+          iconClickFallbackImagesElement,
+          'IconClickFallbackImage'
+        )
+        .forEach((iconClickFallbackImageElement, id) => {
+          icon.iconClickFallbackImages.push({
+            id,
```
ZacharieTFR, line 323: Why adding the `id` attribute ? As there is no reference to it in the VAST specs

docs/api/class-reference.md
```diff
@@ -212,3 +212,18 @@ This object represents a generic Creative. It's used as a parent object for more
...
+
+## survey<a name="survey"></a>
+
+- `value: String`
+- `type: String|null`
+
+## iconClickFallbackImages <a name="iconclickfallbackimages"></a>
+
+- `id: Number`
+- `url: String`
```
ZacharieTFR, line 229: This is missleading as iconClickFallbackImages is an array and the object correspond to the  `iconClickFallbackImage` (singular)
Also, can the url equal to null or will it be an empty string ?

spec/samples/creatives.js
```diff
@@ -46,6 +49,11 @@ export const creatives = `<Creatives>
             <IconClickThrough><![CDATA[http://example.com/linear-clickthrough]]></IconClickThrough>
             <IconClickTracking id='icon-click-1'><![CDATA[http://example.com/linear-clicktracking1]]></IconClickTracking>
             <IconClickTracking id='icon-click-2'><![CDATA[http://example.com/linear-clicktracking2]]></IconClickTracking>
+            <IconClickFallbackImages>
+              <IconClickFallbackImage width="10px" height="10px">
+                <![CDATA[http://adserver.com/fallback.png]]>
+              </IconClickFallbackImage>
+            </IconClickFallbackImages>
```
ZacharieTFR, line 56: We may need a use case with multiple <IconClickFallbackImage> to fully cover use cases

docs/api/class-reference.md
```diff
@@ -212,3 +212,19 @@ This object represents a generic Creative. It's used as a parent object for more
...
+
+- `value: String`
+- `xmlEncoded: String|null`
+
+## survey<a name="survey"></a>
+
+- `value: String`
+- `type: String|null`
+
+## iconClickFallbackImage <a name="iconclickfallbackimage"></a>
```
ZacharieTFR, line 226: To stay coherent with other sections can you put titles in uppercase ? 🙏

src/parser/parser_utils.js
```diff
@@ -338,6 +338,17 @@ function mergeWrapperAdData(unwrappedAd, wrapper) {
       wrapper.blockedAdCategories
     );
   }
+  
+  if (wrapper.creatives?.length) {
+    const wrapperCreativesWithIconsNode = wrapper.creatives.filter(
+      (creative) => creative.icons?.length && !creative.mediaFiles.length
```
ZacharieTFR, line 344: Maybe adding a small comment on why we're filtering out the creative with mediafile can be usefull here

src/parser/vast_parser.js
```diff
@@ -27,6 +27,7 @@ export class VASTParser extends EventEmitter {
   constructor() {
     super();
 
+    this.parseDuration = parserUtils.parseDuration;
```
ZacharieTFR, line 30: One discutions we had about exposing/exporting util functions is that we can either do like you did but then we need to create a new instance of vastParser otherwise the second solution would be to export utils function as a standalone then we could use it without creating a new instance and in addition if we don't use it we could benefit from tree-shaking

docs/api/vast-parser.md
```diff
@@ -355,6 +355,34 @@ vastParser.parseVAST(vastXml, options)
...
+console.log(d); // output -1
+
+```
+
+#### Parameter
+
+- **`duration: String`** - The duration represented as a string.
+
+
+
```
ZacharieTFR, line 385: Too many breaklines here 😅 
```suggestion
```

ZacharieTFR (approved):
LGTM just a small nitpicking

clementFrancon (approved):
LGTM ! Great work !

PR #448, opened 2023-06-12
Fix/Validate URLTemplates before tracking

Lotphy:
### Description

Check and validate tracking urls before requesting them.

### Issue

Provided tracking urls are not always valid urls

### Type
- [ ] Breaking change
- [x] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/util/util.js
+++ b/src/util/util.js
@@ -125,6 +125,25 @@ function extractURLsFromTemplates(URLTemplates) {
 }
 
 /**
+ * Filter URLTemplates elements to keep only valid and safe URL templates.
+ *   To be valid, urls should:
+ *   - have the same protocol as the client
+ *   or
+ *   - be protocol-relative urls
+ *
+ * @param {String|String[]|Object[]} URLTemplates - A string|string[]|object[] of url templates.
+ */
+function filterValidUrlTemplates(URLTemplates) {
+  if (Array.isArray(URLTemplates)) {
+    return URLTemplates.filter(urlTemplate => {
+      const url = urlTemplate.hasOwnProperty('url') ? urlTemplate.url : urlTemplate;
+      return (url.startsWith(location.protocol) || url.startsWith('//'))
+    })
+  }
+  return (URLTemplates.startsWith(location.protocol) || URLTemplates.startsWith('//'));
+}
+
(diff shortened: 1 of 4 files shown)


src/util/util.js
```diff
@@ -124,6 +124,29 @@ function extractURLsFromTemplates(URLTemplates) {
...
 }
 
+/**
+ * Filter URLTemplates elements to keep only valid and safe URL templates.
+ *   To be valid, urls should:
+ *   - have the same protocol as the client
+ *   or
+ *   - be protocol-relative urls
+ *
+ * @param {String|String[]|Object[]} URLTemplates - A string|string[]|object[] of url templates.
```
rumesh, line 134:
```suggestion
 * @param {Array} URLTemplates - An Array of url templates.
```
It should be only this no? As we are expecting an array all the time, and also this is the input value from `trackURLs`

src/util/util.js
```diff
@@ -124,6 +124,29 @@ function extractURLsFromTemplates(URLTemplates) {
...
+    return URLTemplates.filter(urlTemplate => {
+      const url = urlTemplate.hasOwnProperty('url') ? urlTemplate.url : urlTemplate;
+      return isValidUrl(url);
+    })
+  }
+  return isValidUrl(URLTemplates);
+}
+
+function isValidUrl(url) {
+  return (url?.startsWith('http://') || url?.startsWith('https://') || url?.startsWith('//'));
```
rumesh, line 147: Isn't it more performant and also more elegant with only one regex?

Lotphy: Not sure, because this can make http url valid for https hosts, and I'm pretty bad at regex xD.

My first version used window.location but the testing can be limited if we mock the location to https://something for example.
`return (url?.startsWith(window.location.protocol) || url?.startsWith('//'))`

src/util/util.js
Lotphy: Good one. I needed to check after the urls that are not in an array such as survey. As far as I know, this is the only case, and it's not called using trackURLs. Update inc.

ZacharieTFR (approved):

Issue #450, opened 2023-07-03
Issues with adPod containing redirects


radiantmediaplayer:
Hi,

I am trying to load an parse https://www.radiantmediaplayer.com/vast/tags/test1.xml with the following code

```
    const vastClient = new VASTClient();

    vastClient.get('https://www.radiantmediaplayer.com/vast/tags/test1.xml')
      .then(parsedVAST => {
        console.log(parsedVAST);
      })
      .catch(err => {
        // Deal with the error
      });
```

The VAST returns an adPod with 2 Ad and each is a redirect (VASTAdTagURI).
VASTClient however only returns one Ad with parsedVAST in the pod: parsedVAST.ads Array only shows the first ad in the pod, not the second one.
Am I missing something or is it a shortcoming of VASTClient? 

Thanks for looking into it.

radiantmediaplayer:
This seems to be the cause of the issue. When the URIs in VASTAdTagURI are the same as in 
```
<?xml version="1.0" encoding="utf-8"?>
<VAST version="4.0" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="http://www.iab.com/VAST">
  <Ad id="a1" sequence="1">
    <Wrapper>
      <AdSystem>preproll 1</AdSystem>
      <VASTAdTagURI><![CDATA[https://www.radiantmediaplayer.com/vast/tags/inline-linear-1.xml]]></VASTAdTagURI>
    </Wrapper>
  </Ad>
  <Ad id="a2" sequence="2">
    <Wrapper>
      <AdSystem>preproll 2</AdSystem>
      <VASTAdTagURI><![CDATA[https://www.radiantmediaplayer.com/vast/tags/inline-linear-1.xml]]></VASTAdTagURI>
    </Wrapper>
  </Ad>
</VAST>
```
VASTClient only returns 1 Ad, but when the URIs in VASTAdTagURI are different as in 
```
<?xml version="1.0" encoding="utf-8"?>
<VAST version="4.0" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns="http://www.iab.com/VAST">
  <Ad id="a1" sequence="1">
    <Wrapper>
      <AdSystem>preproll 1</AdSystem>
      <VASTAdTagURI><![CDATA[https://www.radiantmediaplayer.com/vast/tags/inline-linear-1.xml]]></VASTAdTagURI>
    </Wrapper>
  </Ad>
  <Ad id="a2" sequence="2">
[...]

PR #452, opened 2023-07-19
Fix/Handle duplicate VASTAdTagURI in AdPod

Rapha0511:
### Description

 the `parentURLs` array was used to store and check if a VASTAdTagURI has been called before, to prevent multiple calls to the same URL. This array has been removed as it is no longer necessary.

Now, wrappers within an AdPod can reference the same VASTAdTagURI. This change is based on the consideration that, in case where a wrapper is calling itself, the fetching process will stop after the `maxWrapperDepth` is reached. This mechanism prevents infinite loops.

### Issue

#450 

### Type
- [ ] Breaking change
- [ ] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/spec/vast_parser.spec.js
+++ b/spec/vast_parser.spec.js
@@ -4,8 +4,6 @@ import { urlFor, fetchXml } from './utils/utils';
 import { util } from '../src/util/util';
 import { parserUtils } from '../src/parser/parser_utils';
 import * as Bitrate from '../src/parser/bitrate';
-import { promises } from 'dns';
-import { expect } from '@jest/globals';
 
 const xml = new DOMParser().parseFromString('<VAST></VAST>', 'text/xml');
 const urlHandlerSuccess = {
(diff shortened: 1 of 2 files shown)


spec/vast_parser.spec.js
```diff
@@ -80,23 +80,32 @@ describe('VASTParser', () => {
...
+      it('applies url filters', (done) => {
         VastParser.URLTemplateFilters = [(url) => url.replace('foo', 'bar')];
 
-        return VastParser.fetchVAST('www.foo.foo').finally(() => {
-          expect(VastParser.parentURLs).toEqual(['www.bar.foo']);
+        const urlHandlerSpy = jest.spyOn(VastParser.urlHandler, 'get');
+
+        VastParser.fetchVAST('www.foo.foo').then(() => {
+          done();
+          expect(urlHandlerSpy).toHaveBeenCalledWith(
```
ZacharieTFR, line 90: You might want to put the done after the expect as Jest do in their documentation https://jestjs.io/docs/asynchronous#callbacks

ZacharieTFR (commented):
LGTM ! 🎉 Just left a small comment about unit tests

ZacharieTFR (approved):
LGTM ! 👍

PR #464, opened 2024-01-05
Add raw XML string to url handler callback

epechuzal:
### Description

Include the raw XML string in the `details` object returned by node URL handler.

We would like to use this for debugging purposes.

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/urlhandlers/node_url_handler.js
+++ b/src/urlhandlers/node_url_handler.js
@@ -34,6 +34,7 @@ function get(url, options, cb) {
         cb(null, xml, {
           byteLength: Buffer.from(data).byteLength,
           statusCode: res.statusCode,
+          rawXml: data,
         });
       });
     });


epechuzal:
@Rapha0511 Thank you! 

I noticed the url handler param and that's what I do locally (my url handler is basically a copy/paste of this one + the raw xml change). It sounds like when the update comes I can just ditch the url handler and go back to using the provided one :) 

Should I close this PR? Sounds like it is redundant given the upcoming changes.

Rapha0511:
Hello !

Your suggested change has been pushed on the PR and you will be able to use the url handler provided by the vast-client on the next release :)
I'm closing this PR since it will be a duplicate change.

Thank you for your contribution :)

Issue #465, opened 2024-01-12
Bug: Fallback not working if top Ad Elements returns empty Vast or fails in case of multiple Ad Elements


shubham-si:
https://github.com/dailymotion/vast-client-js/blob/master/src/parser/vast_parser.js#L461C25-L461C25

`const resolvedAds = util.flatten(unwrappedAds);` default to `[]`  [Link](https://github.com/dailymotion/vast-client-js/blob/master/src/util/util.js#L216)

`if (!resolvedAds && this.remainingAds.length > 0) {`

There seems to be a issue here `!resolvedAds`. Since `!Array<T>` gives `false` always, thus the code within if will not be executed at all.

So If any Ad element (going from top to bottom) fails to fetch xml or return empty vast, it will not fallback to next Ad Element.

```xml
<VAST xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="vast.xsd" version="3.0">
<Ad>
<Wrapper fallbackOnNoAd="true">
<AdSystem/>
<Error>
<![CDATA[ https://error-pixel?[ERRORCODE] ]]>
</Error>
<VASTAdTagURI>
<![CDATA[ ]]>
</VASTAdTagURI>
<Extensions>
<Extension type="waterfall" fallback_index="0"/>
</Extensions>
</Wrapper>
</Ad>
<Ad>
<Wrapper fallbackOnNoAd="true">
<AdSystem/>
<Error>
<![CDATA[ https://error-pixel?[ERRORCODE] ]]>
</Error>
<VASTAdTagURI>
<![CDATA[ does-not-exist ]]>
</VASTAdTagURI>
<Extensions>
<Extension type="waterfall" fallback_index="1"/>
</Extensions>
</Wrapper>
</Ad>
<Ad>
<Wrapper>
<AdSystem/>
<Error>
<![CDATA[ https://error-pixel?[ERRORCODE] ]]>
</Error>
<VASTAdTagURI>
<![CDATA[ https://glomex.github.io/vast-ima-player/linear-ad.xml ]]>
</VASTAdTagURI>
<Extensions>
<Extension type="waterfall" fallback_index="2"/>
</Extensions>
</Wrapper>
</Ad>
</VAST>
```

**Fix:**  `if (resolvedAds.length === 0 && this.remainingAds.length > 0) {`

so that if resolvedAds fails and we have remainingAds, can be proceeded with them.

[...]

PR #466, opened 2024-01-23
fix bug where fallback was not working if first ad element return empty

Rapha0511:
### Description

https://github.com/dailymotion/vast-client-js/blob/master/src/parser/vast_parser.js#L461C25-L461C25

The condition will never be reach since `!resolvedAds` will always be `false`.
So in case the first Ad element of a VAST is empty or it fails, it will never fallback to the next Ad element.

### Issue

If any Ad element (going from top to bottom) fails to fetch xml or return empty vast, it will not fallback to next Ad Element.

[Issue Link](https://github.com/dailymotion/vast-client-js/issues/465
)

### Type
- [ ] Breaking change
- [ ] Enhancement
- [X] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/src/parser/vast_parser.js
+++ b/src/parser/vast_parser.js
@@ -458,7 +458,7 @@ export class VASTParser extends EventEmitter {
     return Promise.all(resolveWrappersPromises).then((unwrappedAds) => {
       const resolvedAds = util.flatten(unwrappedAds);
 
-      if (!resolvedAds && this.remainingAds.length > 0) {
+      if (!resolvedAds.length && this.remainingAds.length > 0) {
         const remainingAdsToResolve = this.remainingAds.shift();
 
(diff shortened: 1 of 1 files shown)


src/parser/vast_parser.js
```diff
@@ -458,7 +458,7 @@ export class VASTParser extends EventEmitter {
     return Promise.all(resolveWrappersPromises).then((unwrappedAds) => {
       const resolvedAds = util.flatten(unwrappedAds);
 
-      if (!resolvedAds && this.remainingAds.length > 0) {
```
shubham-si, line 461: looks good to me

```text
!0 = true
!(non zero length) = false
```

shubham-si:
@Rapha0511 when we are planning to release this fix

PR #467, opened 2024-01-23
Upgraded fetching method, added error logging, and code Improvment

Rapha0511:
### Description

This pull request is regroup 2 branches :
- [upgrade url handler to fetch](https://github.com/dailymotion/vast-client-js/tree/urlHandler-upgrade-to-fetch) to upgrade the url handler from XMLHttpRequest to fetch.
Now, there is only one fetching method for browser and Node environment

-  [code improvment](https://github.com/dailymotion/vast-client-js/tree/code-improvment) to improve some code for a better maintainability and readability

- it also includes more error logs for  better debugging

### Type
- [ ] Breaking change
- [X] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling


src/urlhandlers/xhr_url_handler.js
```diff
@@ -1,67 +1,90 @@
...
+    const error = handleError(response);
+    if (error) {
+      return { error: new Error(error), statusCode: response.status };
+    }
+    return handleResponse(response);
   } catch (error) {
-    cb(new Error('XHRURLHandler: Unexpected error'));
+    return {
+      error,
+      statusCode: error.name === 'AbortError' ? 408 : null,
```
ZacharieTFR, line 83: Instead of null shouldn't we return response.status ?

src/util/util.js
```diff
@@ -125,22 +125,34 @@ function extractURLsFromTemplates(URLTemplates) {
...
-  }
-  return isValidUrl(URLTemplates);
+function filterUrlTemplates(URLTemplates) {
+  return URLTemplates.reduce(
+    (acc, urlTemplate) => {
+      const url = urlTemplate.hasOwnProperty('url')
+        ? urlTemplate.url
+        : urlTemplate;
+
+      let urlVerification = isValidUrl(url);
```
ZacharieTFR, line 149: IMO this variable is not necessary

src/util/util.js
```diff
@@ -125,22 +125,34 @@ function extractURLsFromTemplates(URLTemplates) {
...
-      return isValidUrl(url);
-    })
-  }
-  return isValidUrl(URLTemplates);
+function filterUrlTemplates(URLTemplates) {
+  return URLTemplates.reduce(
+    (acc, urlTemplate) => {
+      const url = urlTemplate.hasOwnProperty('url')
+        ? urlTemplate.url
+        : urlTemplate;
```
ZacharieTFR, line 147: In which case can we have an object instead ? 
```suggestion
      const url = urlTemplate.url || urlTemplate;
```

src/vast_tracker.js
```diff
@@ -151,6 +151,10 @@ export class VASTTracker extends EventEmitter {
   setDuration(duration) {
     // check if duration is a valid time input
     if (!util.isValidTimeValue(duration)) {
+      this.emit('TRACKER-error', {
```
ZacharieTFR, line 154: Can you update the documentation for thoses new errors ?

src/vast_tracker.js
```diff
@@ -180,8 +184,14 @@ export class VASTTracker extends EventEmitter {
   setProgress(progress, macros = {}) {
     // check if progress is a valid time input
     if (!util.isValidTimeValue(progress) || typeof macros !== 'object') {
+      this.emit('TRACKER-error', {
+        message: 'setProgress parameters has the wrong type.',
```
ZacharieTFR, line 188:
```suggestion
        message: 'setProgress parameters are not valid.',
```

src/vast_tracker.js
```diff
@@ -151,6 +151,10 @@ export class VASTTracker extends EventEmitter {
   setDuration(duration) {
     // check if duration is a valid time input
     if (!util.isValidTimeValue(duration)) {
+      this.emit('TRACKER-error', {
+        message: 'the duration provided is not valid',
```
ZacharieTFR, line 155:
```suggestion
        message: 'setDuration parameter duration is not valid',
```

src/vast_tracker.js
```diff
@@ -327,6 +355,10 @@ export class VASTTracker extends EventEmitter {
    */
   setSkipDelay(duration) {
     if (!util.isValidTimeValue(duration)) {
+      this.emit('TRACKER-error', {
+        message: 'setSkipDelay parameter does not have a valid value',
+        parameter: duration,
```
ZacharieTFR, line 360: I don't find having multiple property for the tracker-error convenient, because it error prone and then to use we multiply the properties to check for user. If I sum up I see that sometimes you will have only `message` sometime "parameter" singular and `parameters` plural and the contained values doesn't always have the same type. Maybe simply adding them to the message is enough

src/vast_tracker.js
```diff
@@ -339,6 +371,10 @@ export class VASTTracker extends EventEmitter {
    */
   trackImpression(macros = {}) {
     if (typeof macros !== 'object') {
+      this.emit('TRACKER-error', {
+        message: 'trackImpression parameter has the wrong type',
```
ZacharieTFR, line 375: The message can be more explicit in this case and next ones below
```suggestion
        message: 'trackImpression given macros has the wrong type',
```

src/urlhandlers/xhr_url_handler.js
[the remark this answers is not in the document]
ZacharieTFR: Also remove the folder

spec/fetcher.spec.js
```diff
@@ -162,12 +160,15 @@ describe('Fetcher', () => {
...
-        });
-        return expect(result).rejects.toEqual(new Error('timeout'));
+        let result = fetcher
+          .fetchVAST({
+            url: url,
+            maxWrapperDepth: 5,
+            emitter: () => {},
+          })
+          .catch(() => {
+            return expect(result).rejects.toEqual('error');
```
ZacharieTFR, line 170: Is the error type different here ? Because it should be iso with was we had before in term of behavior

src/util/util.js
```diff
@@ -125,22 +125,34 @@ function extractURLsFromTemplates(URLTemplates) {
...
+
+      let urlVerification = isValidUrl(url);
+
+      urlVerification ? acc.validUrls.push(url) : acc.unvalidUrls.push(url);
+      return acc;
+    },
+    { validUrls: [], unvalidUrls: [] }
```
ZacharieTFR, line 154: Typo 
```suggestion
    { validUrls: [], invalidUrls: [] }
```

spec/util.spec.js
```diff
@@ -374,4 +374,48 @@ describe('util', function () {
...
+        },
+      ];
+
+      const { validUrls, unvalidUrls } = util.filterUrlTemplates(urlsToFilter);
+      expect(validUrls.length).toBe(3);
+      expect(unvalidUrls.length).toBe(2);
+    });
+  });
+
+  describe('#isValidUrls', function () {
```
ZacharieTFR, line 408: Typo
```suggestion
  describe('#isValidUrl', function () {
```

spec/util.spec.js
```diff
@@ -374,4 +374,48 @@ describe('util', function () {
...
+    it('should return true if the url is valid', () => {
+      const url = 'http://example.com';
+      const validUrl = util.isValidUrl(url);
+      expect(validUrl).toBe(true);
+    });
+
+    it('should return false if the url is not valid', () => {
+      const url = 'example.com';
+      const validUrl = util.isValidUrl(url);
```
ZacharieTFR, line 417: 😉 
```suggestion
      const invalidUrl = util.isValidUrl(url);
```

spec/vast_client.spec.js
```diff
@@ -1,19 +1,37 @@
 import { VASTClient } from '../src/vast_client';
-import { nodeURLHandler } from '../src/urlhandlers/node_url_handler';
-import { urlFor } from './utils/utils';
-
-const wrapperMultipleAdsVastUrl = urlFor('wrapper-multiple-ads.xml');
-const emptyVastUrl = urlFor('empty-no-ad.xml');
+import { readFile } from 'fs/promises';
+import { afterAll } from '@jest/globals';
```
ZacharieTFR, line 3: Why the import is needed here?

src/fetcher.js
```diff
@@ -76,47 +76,46 @@ export class Fetcher {
...
-            resolve(xml);
-          }
+        if (data.error) {
+          reject(data.error);
+        } else {
+          resolve(data.xml);
         }
-      );
+      } catch (error) {
+        console.error(error);
```
ZacharieTFR, line 117: We should have console logs, you should reject instead

ZacharieTFR (changes requested):

src/parser/ad_parser.js
```diff
@@ -22,32 +22,42 @@ export function parseAd(
...
+    .filter((childNode) =>
+      ['inline', 'wrapper'].includes(childNode.nodeName.toLowerCase())
+    )
+    .filter(
+      (adType) =>
+        !(
+          adType.nodeName.toLowerCase() === 'wrapper' &&
+          followAdditionalWrappers === false
+        )
+    );
```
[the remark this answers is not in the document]
Rapha0511: Since followAdditionalWrapper can be undefined, i think the filter will be more precise if we do it this way.

const filteredChildNodes = childNodes.filter((childNode) => {
const childNodesToLowerCase = childNode.nodeName.toLowerCase();
return (
childNodesToLowerCase === 'inline' ||
(followAdditionalWrappers !== false &&
childNodesToLowerCase === 'wrapper')
);
});
`

spec/fetcher.spec.js
```diff
@@ -122,13 +124,21 @@ describe('Fetcher', () => {
...
             expect(urlHandlerSpy).toHaveBeenCalledWith(
               expectedUrl,
-              expect.anything(),
               expect.anything()
             );
           });
       });
 
       it('should emits VAST-resolving and VAST-resolved events with a filtered url', () => {
+        fetcher.setOptions({});
```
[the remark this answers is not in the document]
Rapha0511: When using the vastclient we never have to use `fetcher.setOptions` directly.

`fetcher.setOptions` is called when using `vastclient.get` or `vastclient.parseVAST` but we never have to manipulate it directly.

This is just for testing purposes. 

The reason i called `fetcher.setOptions()` with empty params is to be able to reset the options and  mock the `fetch` function used by the  `urlHandler` and have a realistic error throw by the vastclient.

I hope this makes things clearer!

ZacharieTFR (approved):

clementFrancon (approved):
LGTM ! Great work !

PR #474, opened 2024-07-18
AbortHandler's timeout should always be cleared

olafbuitelaar:
the timeout should always be cleared even if an error occurred within the promise

### Description

When you provide a custom urlHandler and an error occurs here-in it could be possible the abortHandlers timeout isn't cleared, causing the abortHandler to trigger even if the request is done.

### Type
- [ x] Fix

### DIFF
--- a/src/fetcher/url_handler.js
+++ b/src/fetcher/url_handler.js
@@ -66,8 +66,10 @@ async function get(url, options) {
       ...options,
       signal: controller.signal,
       credentials: options.withCredentials ? 'include' : 'omit',
+    })
+    .finally(()=>{
+      clearTimeout(timer);
     });
-    clearTimeout(timer);
 
     const error = handleError(response);
     if (error) {


src/fetcher/url_handler.js
```diff
@@ -66,8 +66,10 @@ async function get(url, options) {
       ...options,
       signal: controller.signal,
       credentials: options.withCredentials ? 'include' : 'omit',
+    })
+    .finally(()=>{
+      clearTimeout(timer);
```
Rapha0511, line 71: Can this be done in a `finally` block after the `catch` ? since `.finally()` is usually used in a Promise chain

Rapha0511 (commented):
Hello!
Thank you for reporting this issue! 
I left some comment.

src/fetcher/url_handler.js
olafbuitelaar: the `await` isn't that different that a promise chain, only different syntax. if you would want to move it to the try..catch..finally. the `const timer` needs to be moved outside the try block since it's block scoped. probably all together less neat.

olafbuitelaar:
any update on this? i hoped it would have been included in 6.0.1

Rapha0511 (approved):

clementFrancon (approved):

olafbuitelaar:
when is it planned to publish this?

ZacharieTFR:
> when is it planned to publish this?

It's ready to use! 🎉 Available in `6.0.2` https://github.com/dailymotion/vast-client-js/releases/tag/6.0.2

Issue #476, opened 2024-09-21
VASTClient not throwing error or emitting wrong error code for invalid wrapper VAST


ccampbell5:
Hi, I'm using version 6.0.1 and noticing inconsistent behavior with the VASTClient in how it processes wrapper VASTs that contain errors. The following sample code contains two wrapper VASTs; expected vs actual behavior is indicated in comments. (Just swap the `test` constant.) Also attaching the package.json file. [package.json](https://github.com/user-attachments/files/17085876/package.json)

1. The `expects303Error` VAST is a two-layer wrapper that eventually points to an empty VAST document: `<VAST xsi:noNamespaceSchemaLocation="vast.xsd" version="3.0"/>`. According to the [docs](https://github.com/dailymotion/vast-client-js/blob/5d148f75770aaca03edf64c0739d9523fad06293/docs/api/vast-parser.md#error-tracking), the error code for "No VAST response after one or more Wrappers" is 303. However, no 'VAST-error' event is emitted for this VAST. 
2. The `expectsThrowError` wrapper VAST points to an unsupported 1.0 VAST document; in this case, my understanding is that the VASTClient's `parseVAST` method should return a rejected Promise since the root element is not 'VAST'. However, the resulting promise is not rejected, and the parser emits a 'VAST-error' with the 301 (timeout) code. (If you follow the VASTAdTagURI link and copy the VAST 1.0 XML from there, then pass it to `parseVAST`, then the expected rejected Promise is returned.)

```
import { VASTClient } from "@dailymotion/vast-client";
import xmlDom from "@xmldom/xmldom";
import fetch from "node-fetch";


const client = new VASTClient();

client.getParser().on("VAST-error", (err) => {
  console.log("VAST-error: ", err);
});

const getXML = (text) => {
  const parser = new xmlDom.DOMParser({
    errorHandler: (level, msg) => {
      throw new Error(`[Level: ${level}]: ${msg}`);
    },
  });

  const xml = parser.parseFromString(text, "text/xml");
  const err = xml
    ?.querySelector?.("parsererror")
    ?.querySelector?.("div")?.textContent;

  if (err) {
    throw new Error(err);
  }

  return xml;
};

// Expected:
// Should emit a 303 error since
[...]

PR #481, opened 2024-11-12
Add support for non supported vast error

ZacharieTFR:
### Description

- vast client will now trigger error code `102` `VAST version of response not supported.` if it parse a outdated VAST version (only 1.0 for now)
- Fix error `303` `No VAST response after one or more Wrappers` not being triggered when a wrapper containing a `<Creative>` without mediafile (e.g only with a `<CompanionAds>` inside) does not result in a VAST response.

### Issue

#476 

### Type
- [ ] Breaking change
- [ ] Enhancement
- [X] Fix
- [X] Documentation
- [ ] Tooling


docs/api/1.x/parser.md
```diff
@@ -7,6 +7,8 @@ Whenever an error occurs during the VAST parsing, the parser will call on his ow
 
 * VAST error `101`: VAST schema validation error.
 
+* VAST error `102`: VAST version of response not supported.
```
clarissedmn, line 10:
```suggestion
* VAST error `102`: VAST response version not supported.
```

clarissedmn (approved):
LGTM 🎉

clementFrancon (approved):
LGTM !

docs/api/1.x/parser.md
ZacharieTFR: Actually I edited the wrong file, this documentation is for old versions 1.x. I moved it to the correct one. The error description comes from IAB https://wiki.iabtechlab.com/index.php?title=VAST_Error_Code_Troubleshooting_Matrix

Issue #360, opened 2020-06-08
Opinion on allowing unordered ad sequences on Ad Pods for resolveAll method


ZacharieTFR:
As mentionned in https://github.com/dailymotion/vast-client-js/issues/359#issuecomment-632107969 for a Ad Pod that have multiple Ad elements with unique sequence attributes inside a VAST but are not in ascending order the resolveAll() method will return only the first Ad element.

Meaning that resolveAll() return the first Ad element found in VAST for any VAST that contain an Ad Pod with:
- sorted in descending order (e.g for Ad sequence3>Ad sequence2>Ad sequence1 it return Ad sequence3)
- non sorted Ad elements (e.g for Ad sequence1>Ad sequence3>Ad sequence2 >it return Ad sequence1)

Note: When the VAST is parsed using basic methods (vastClient.get, vastParser.parseVAST, vastParser.getAndParseVAST etc..) all VAST elements are correctly parsed and returned inside a JSON object. If an Ad pod is unordered the parsing result with return parsed elements in the same order. 

Please comment this issue if you have an express need for it. If so we will add the feature. In the meanwhile feel free to contribute by creating a pull request to add the support of this feature! 💪

PR #487, opened 2025-01-24
Fix AdPods parsing behavior with resolveAll to false

ZacharieTFR:
### Description
Previously, when a VAST contained an adPod with standalone ads, if a standalone ad appeared first in the VAST order, it would be parsed first, causing the adPod to not be resolved. As a result, the Player couldn't be aware that the VAST contained an adPod. According to IAB specifications, the adPod should take priority and play before any standalone ads. This PR addresses the issue by ensuring that the adPod is resolved first, followed by the standalone ads, when `getNextAds` is called.

### Issue
[#360](https://github.com/dailymotion/vast-client-js/issues/360)

### Type
- [ ] Breaking change
- [x] Enhancement
- [x] Fix
- [ ] Documentation
- [ ] Tooling

### DIFF
--- a/spec/vast_client.spec.js
+++ b/spec/vast_client.spec.js
@@ -118,20 +118,17 @@ describe('VASTClient', () => {
           optionsWithNoResolveAll
         );
       });
-      it('returns first ad parsed', () => {
-        expect(res).toEqual({
-          ads: expect.any(Array),
-          errorURLTemplates: [],
-          version: '4.3',
-        });
+      it('returns ad pod parsed', () => {
         expect(res.ads).toHaveLength(2);
+        expect(res.ads[0].sequence).toBe('1');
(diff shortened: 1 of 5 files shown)


spec/vast_client.spec.js
```diff
@@ -161,7 +158,7 @@ describe('VASTClient', () => {
           expect(res.ads).toHaveLength(3);
         });
 
-        it('resolves only next ad if requested', async () => {
+        fit('resolves only next ad if requested', async () => {
```
clarissedmn, line 161: Perhaps we can create a ticket to add an ESLint rule to prevent fit from being pushed to production?
```suggestion
        it('resolves only next ad if requested', async () => {
```

clarissedmn:
After removing `fit` from the tests, it looks like two tests are failing

clementFrancon (approved):

PR #504, opened 2026-03-20
feat: Add wrapper chain id to vast resolution events

clementFrancon:
### Description

Since we resolve wrappers in parallel when resolveAll option is set to true it's not very intuitive to follow the wrapper resolution while troubleshooting or monitoring as the resolve events are mixed together. 
I added a simple increment as a "wrapper chain" id so we can follow them more effectively. 

### Type
- [ ] Breaking change
- [x] Enhancement
- [ ] Fix
- [ ] Documentation
- [ ] Tooling


clarissedmn (approved):
Seems good !!

